PR #10, opened 2021-12-22
Add decodeFirstPage

sz-piotr:


packages/encoding/src/readFundingIndices.ts
```diff
@@ -0,0 +1,16 @@
+import { ByteReader } from './ByteReader'
+import { MIN_INT } from './constants'
+import { decodeAssetId } from './decodeAssetId'
+import { FundingIndex } from './OnChainData'
+
+export function readFundingIndices(reader: ByteReader) {
+  const fundingIndicesLength = reader.readNumber(32)
+  const indices: FundingIndex[] = []
+  for (let i = 0; i < fundingIndicesLength; i++) {
+    reader.skip(17)
```
hasparus, line 10: `reader.skip` mutates, so we need to call these functions in a set order, right?

packages/encoding/src/readState.ts
```diff
@@ -0,0 +1,8 @@
+import { ByteReader } from './ByteReader'
+import { decodeState } from './decodeState'
+
+export function readState(reader: ByteReader) {
```
hasparus, line 4: Do we have some docs about this data layout that we could mention in the readme/comments?

packages/encoding/test/decodeAssetId.test.ts
```diff
@@ -1,6 +1,6 @@
 import { expect } from 'chai'
 
-import { decodeAssetId, encodeAssetId } from '../src/assetId'
+import { decodeAssetId, encodeAssetId } from '../src/decodeAssetId'
```
hasparus, line 3: nit: I'm not very attached to file names personally, but `encodeAssetId` is now the living in the shadow of its inverse function :P

packages/encoding/test/decodeUpdates.test.ts
```diff
@@ -0,0 +1,97 @@
...
+      DecodingError,
+      'Went out of bounds'
+    )
+  })
+
+  it('decodes multiple entries and positions', () => {
+    const writer = new ByteWriter()
+      .writeNumber(2, 32) // 2 entries
+      .writeNumber(3, 32) // 3 indices
+      .writePadding(17)
```
hasparus, line 21: Very neat chaining.

hasparus:
Whenever a number is signed, we add MIN_INT, right? How do we know which ones are signed? Do we have some docs on this encoding to refer to?

hasparus (approved):

packages/encoding/src/readFundingIndices.ts
sz-piotr: Yup

packages/encoding/test/decodeAssetId.test.ts
sz-piotr: Oh you're right. I'm going to change the file name

sz-piotr:
The encoding is defined in the python codebase we've been given

PR #12, opened 2022-01-05
Add database and PositionUpdateRepository

hasparus:
### What's changed

- Added `DatabaseService`
- Added `PositionUpdateRepository` and `001_positions.ts` migration.
- Told node-postgres BigInts already exist in Node.


packages/frontend/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "lib": ["es2020", "DOM"],
```
hasparus, line 4: @sz-piotr I'm not quite sure why it worked before, but this has been `yarn typecheck` for me.

https://github.com/l2beat/dydx-state-explorer/runs/4726819953?check_suite_focus=true

image

.eslintrc.json
```diff
@@ -16,7 +16,8 @@
     "@typescript-eslint/no-extra-semi": "off",
     "@typescript-eslint/no-empty-function": "off",
     "simple-import-sort/imports": "error",
-    "simple-import-sort/exports": "error"
+    "simple-import-sort/exports": "error",
+    "@typescript-eslint/no-namespace": ["warn", { "allowDeclarations": true }]
```
sz-piotr, line 20: warn -> error

why is this check needed?

packages/backend/src/model/AssetBalances.ts
```diff
@@ -0,0 +1,43 @@
+interface AssetBalance {
+  assetId: string
+  balance: bigint
+}
+
+export class AssetBalances extends Array<AssetBalance> {
```
sz-piotr, line 6: I am not sure if this is the way I'd go about solving this problem.

Why not have toRow and toRecord stringify and parse the values and not have a special type that extends Array? Extending Array is problematic, because all methods return regular arrays not your custom type.

packages/backend/src/config/config.testing.ts
```diff
@@ -12,5 +13,6 @@ export function getTestConfig(): Config {
       format: 'json',
     },
     port: 1337,
+    databaseUrl: getEnv('TEST_DB_URL', 'xXTestDatabaseUrlXx'),
```
sz-piotr, line 16: Improve the name, as the other comment suggests

sz-piotr (changes requested):

.eslintrc.json
hasparus: It's not needed IMHO. It's `error` by default without exemption for declarations, by I wanted to use `export declare namespace` to attach types to a class with declaration merging.

packages/backend/test/peripherals/database/setup.ts
```diff
@@ -0,0 +1,22 @@
+import { getConfig } from '../../../src/config'
+import { DatabaseService } from '../../../src/peripherals/database/DatabaseService'
+
+export function setupDatabaseTestSuite() {
+  const config = getConfig('test')
+  const knex = DatabaseService.createKnexInstance(config.databaseUrl)
+  const skip = config.databaseUrl === 'xXTestDatabaseUrlXx'
```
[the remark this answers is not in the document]
hasparus: Fair point.

image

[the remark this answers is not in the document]
hasparus: Renamed.

If we'd like to improve upon this more, we could consider something like [`pg-mem`](https://github.com/oguimbal/pg-mem)., but since we run these tests on CI anyway, I feel skipping makes sense.

packages/backend/src/config/config.testing.ts
hasparus: What do you feel about this import I just introduced?

packages/backend/src/model/AssetBalances.ts
hasparus: > Extending Array is problematic, because all methods return regular arrays not your custom type.

I remember that was the case in old JavaScript, but I don't think it's the case anymore.

**Node 14**

image

**Browser (Brave)**

image

hasparus: The root cause of the problem is an inconstistency on Knex and `node-postgres` side.

We need to pass a stringified JSON to the database — https://knexjs.org/#Schema-jsonb.
But, when we receive it from the database, we get the JSON object alraedy _partially_ parsed with `JSON.parse`.

After receiving it, I'm turning all `balance` properties into `bigint` — I'm not quite sure it's needed, but I wanted to preserve the types you wrote in `OnChainData.ts`.

hasparus: I can inline this logic in `toRow` and `toRecord` if you'd prefer.

Another question I have is: Would we ever want to filter or query position updates based on an asset balance? Should this maybe become a table in the database instead?

packages/backend/src/config/config.testing.ts
sz-piotr: I like it, makes it clear where the value is coming from

packages/backend/src/model/AssetBalances.ts
sz-piotr: There is also the issue of TypeScript handling of the type:
image

.eslintrc.json
```diff
@@ -16,7 +16,8 @@
     "@typescript-eslint/no-extra-semi": "off",
     "@typescript-eslint/no-empty-function": "off",
     "simple-import-sort/imports": "error",
-    "simple-import-sort/exports": "error"
+    "simple-import-sort/exports": "error",
+    "@typescript-eslint/no-namespace": ["error", { "allowDeclarations": true }]
```
sz-piotr, line 20: I am not sure if this is still necessary. Worst case we can `// eslint-ignore` in one place

packages/backend/src/peripherals/database/DatabaseService.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+}
+
+/**
+ * node-postgres returns bigints as strings since 2013.
+ * @see https://github.com/brianc/node-postgres/pull/353
+ * We set a parser here, to inform pg-types BigInt is supported in Node :)
+ * `20` is the id of 8-bytes {@link pgTypes.TypeId.INT8}
+ */
+pgTypes.setTypeParser(20, 'text', BigInt)
```
sz-piotr, line 59: We discussed moving this to `createKnexInstance`

packages/backend/src/peripherals/database/PolyglotMigrationSource.ts
```diff
@@ -0,0 +1,53 @@
...
+    this.fsSource = new FsMigrations(
+      migrationsDirectory,
+      sortDirsSeparately,
+      loadExtensions
+    )
+  }
+
+  async getMigrations(_loadExtensions: string[]): Promise<unknown[]> {
+    const xs = await this.fsSource.getMigrations(this.loadExtensions)
+    console.log(xs.length)
```
sz-piotr, line 42: Unnecessary

packages/backend/src/peripherals/database/types.ts
```diff
@@ -0,0 +1,34 @@
...
+    asset_id: string
+    balance: string
+  }
+
+  interface Tables {
+    position_updates: PositionUpdateRow
+  }
+}
+
+export interface Repository<TRecord> {
```
sz-piotr, line 22: I am unsure if I like this or not, especially with `addOrUpdate`. Let's keep it for now, I am curious to see how this scales for different repositories.

packages/backend/test/peripherals/database/PositionUpdateRepository.test.ts
```diff
@@ -0,0 +1,141 @@
...
+        fundingTimestamp: BigInt(1621947600),
+        balances: [],
+      },
+      {
+        positionId: BigInt(2),
+        publicKey:
+          '0x015e8410e93e5c90b1bf7a393874f68f4cc7f477c5d742daac5c5a112e672d61',
+        collateralBalance: BigInt(12280156418787),
+        fundingTimestamp: BigInt(1621958400),
+        balances: [
```
sz-piotr, line 56: We can probably do away with having just 3 values here, one positive, one negative and one zero.

sz-piotr (changes requested):
I think a simple readme that tells how to set up the database locally (similar to L2BEAT's one) will be a nice addition

packages/backend/src/peripherals/database/PolyglotMigrationSource.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    this.fsSource = new FsMigrations(
+      migrationsDirectory,
+      sortDirsSeparately,
+      loadExtensions
+    )
+  }
+
+  async getMigrations(_loadExtensions: string[]): Promise<unknown[]> {
+    const xs = await this.fsSource.getMigrations(this.loadExtensions)
```
sz-piotr, line 41: This line is still unnecessary

sz-piotr (changes requested):

packages/backend/src/peripherals/database/PolyglotMigrationSource.ts
hasparus: 🤦‍♂️ sorry!

sz-piotr (approved):

PR #13, opened 2022-01-06
Add wallaby.js configuration file

hasparus:
### What's changed

- Rename config.test.ts to config.testing.ts
- Add WallabyJS configuration file

### Relevant links

- https://wallabyjs.com/


wallaby.js
```diff
@@ -0,0 +1,22 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+require("dotenv").config();
+
+module.exports = function (_wallaby) {
+  const packageNames = ["backend", "encoding", "crypto", "frontend"];
+  const packages =
+    packageNames.length === 1 ? packageNames[0] : `{${packageNames.join(",")}}`;
```
sz-piotr, line 7:
```suggestion
    packageNames.join(",");
```

wallaby.js
```diff
@@ -0,0 +1,22 @@
...
+  const directories = `./packages/${packages}/{src,test}/**/`;
+  const sourceFiles = directories + "*.{ts,mjs,cjs,js,json,env}";
```
sz-piotr, line 9:
```suggestion
  const sourceFiles = directories + "*.{ts,tsx,mjs,cjs,js,jsx,json,env}";
```

wallaby.js
```diff
@@ -0,0 +1,22 @@
...
+  const testFiles = directories + "*.test.ts";
```
sz-piotr, line 10:
```suggestion
  const testFiles = directories + "*.test.{ts,tsx}";
```

wallaby.js
```diff
@@ -0,0 +1,22 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+require("dotenv").config();
```
sz-piotr, line 2: Why is this necessary?

sz-piotr (changes requested):

wallaby.js
hasparus: Your suggestion drops `{` and `}`.

If there is more than one directory name in the array we want `"{one,two}"`, otherwise we just want `"one", because `"{one}"` is illegal.

wallaby.js
hasparus: The file itself or `require("dotenv").config()`?

hasparus: If you're asking about `dotenv`, I wanted to let Wallaby know about `process.env.TEST_DB_URL`.

sz-piotr: The require call

wallaby.js
sz-piotr: Oh! I didn't notice the brackets. Then the code is correct

PR #14, opened 2022-01-07
Move encoding/src/data to backend package and save positions to db

hasparus:


packages/backend/src/index.ts
```diff
@@ -3,15 +3,21 @@ import { getConfig } from './config'
...
 }
 
 function onError(e: unknown) {
   console.error(e)
   process.exit(1)
 }
+
+// Public API exposed for encoding tests
+
+export { getOnChainData } from './peripherals/onchain'
```
sz-piotr, line 23: Maybe we should do it differently. Instead of having `src/index.ts` we could have `src/main.ts` and `src/lib.ts`. Then we'd set package.json's `main` to `lib.js`

sz-piotr (changes requested):

PR #20, opened 2022-01-11
Add SyncScheduler

hasparus:


packages/backend/src/core/SyncScheduler.ts
```diff
@@ -0,0 +1,83 @@
...
+import { BlockNumber, BlockRange } from '../peripherals/ethereum/types'
+import { getBatches } from '../tools/getBatches'
+import { JobQueue } from '../tools/JobQueue'
+import { Logger } from '../tools/Logger'
+import { DataSyncService } from './DataSyncService'
+import { SafeBlockService } from './SafeBlockService'
+
+// block of the first verifier deploy
+const EARLIEST_BLOCK = 11813207
+const BATCH_SIZE = 500
```
hasparus, line 11:
```suggestion
/** max synced blockRange length */
const BATCH_SIZE = 500
```

packages/backend/src/core/SyncScheduler.ts
```diff
@@ -0,0 +1,83 @@
+import { KeyValueStore } from '../peripherals/database/KeyValueStore'
+import { BlockNumber, BlockRange } from '../peripherals/ethereum/types'
+import { getBatches } from '../tools/getBatches'
+import { JobQueue } from '../tools/JobQueue'
+import { Logger } from '../tools/Logger'
+import { DataSyncService } from './DataSyncService'
+import { SafeBlockService } from './SafeBlockService'
+
+// block of the first verifier deploy
```
hasparus, line 9:
```suggestion
/** block of the first verifier deploy */
```

packages/backend/src/tools/Logger.ts
```diff
@@ -22,6 +22,7 @@ export class Logger {
   constructor(private options: LoggerOptions) {}
 
   static SILENT = new Logger({ logLevel: LogLevel.NONE, format: 'pretty' })
+  static DEBUG = new Logger({ logLevel: LogLevel.DEBUG, format: 'pretty' })
```
hasparus, line 25: I'm not 100% sure it's needed here, but it was very useful for testing. I can move it to `test` directory if you'd like @sz-piotr.

sz-piotr: I think it's a nice change.

packages/backend/src/core/SyncScheduler.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  private async sync(blockRange: BlockRange) {
+    this.logger.info({ method: 'sync', blockRange })
+    await this.dataSyncService.sync(blockRange)
+    await this.setLastBlockNumberSynced(blockRange.to)
+  }
+
+  private async getLastBlockNumberSynced(): Promise<BlockNumber> {
+    const valueInDb = await this.kvStore.get('lastBlockNumberSynced')
```
sz-piotr, line 76: Maybe the KeyValueStore should expose methods like `getNumber(key, default)` and `setNumber(key, value). Then those helpers wouldn't be needed.

packages/backend/src/config/config.testing.ts
```diff
@@ -16,5 +16,22 @@ export function getTestConfig(): Config {
     },
     port: 1337,
     databaseUrl: getEnv('TEST_DB_URL', __SKIP_DB_TESTS__),
+    jsonRpcUrl: getEnv('TEST_JSON_RPC_URL', 'http://localhost:8545'),
+    core: throwsOnUnexpectedAccess({
```
sz-piotr, line 20: I like this, but it should probably enforce that all the keys from `core` are present.

sz-piotr (changes requested):

packages/backend/src/config/config.testing.ts
sz-piotr: My bad, it actually is typesafe

packages/backend/src/core/SyncScheduler.ts
hasparus: I have mixed feelings about this. _KeyValueStore_ cannot enforce that the value under `key` is a number, so it would always be possible to "crash" by calling `getNumber("this-is-actually-a-stringified-object")`.

sz-piotr: This is already a problem with the current design, it just propagates an erroneous value instead of crashing

hasparus: From a perspective of `kvStore`, all values are just strings (or undefined) so the returned value is never "erroneous".
 
As _SyncScheduler_ is the only place where `store["lastBlockNumberSynced"]` is written and read it's also the place where it's stringified and parsed.

**I can move these two methods to a separate class if you feel that parsing and stringifying that block number shouldn't be _SyncScheduler_'s responsibility.** What do you think about it?

sz-piotr: Yeah. Makes sense. So we'd have a `SyncStatusRepository` that uses `kvStore` and has `getLastBlockNumberSynced` and `setLastBlockNumberSynced`.

packages/backend/src/peripherals/database/SyncStatusRepository.ts
```diff
@@ -0,0 +1,27 @@
...
+const EARLIEST_BLOCK = 11813207
+
+export type Store = Pick<KeyValueStore<'lastBlockNumberSynced'>, 'get' | 'set'>
+
+export class SyncStatusRepository {
+  private readonly earliestBlock: number
+
+  constructor(
+    private readonly store: Store,
+    options: { earliestBlock?: number } = {}
```
sz-piotr, line 13: why not?
```ts
    private readonly earliestBlock = EARLIEST_BLOCK,
```

sz-piotr (changes requested):

packages/backend/src/peripherals/database/SyncStatusRepository.ts
hasparus: While refactoring SyncScheduler tests I noticed that having two consecutive numbers in constructor args is a bit error prone.


We had earliest block and batchSize, both with default values, so `new SyncScheduler(...objects, 1)` meant `earliestBlock=1, batchSize=500`. After the change it meant `batchSize=1`.

sz-piotr: Makes sense. Let's keep it this way then

packages/backend/test/peripherals/database/SyncStatusRepository.test.ts
```diff
@@ -0,0 +1,51 @@
...
+          key = k
+          return '20'
+        },
+      })
+    )
+
+    const actual = await repository.getLastBlockNumberSynced()
+
+    expect(actual).to.eq(20)
+    expect(key).to.eq('lastBlockNumberSynced')
```
sz-piotr, line 39: You can actually move this expect to the first line of the `async get(k) {` function so that there is no need for `let key = ''`

sz-piotr (changes requested):

packages/backend/test/peripherals/database/SyncStatusRepository.test.ts
hasparus: What if `get` is never called?

packages/backend/src/peripherals/database/SyncStatusRepository.ts
```diff
@@ -0,0 +1,27 @@
...
+  ) {
+    this.earliestBlock = options.earliestBlock || EARLIEST_BLOCK
+  }
+
+  async getLastBlockNumberSynced(): Promise<number> {
+    const valueInDb = await this.store.get('lastBlockNumberSynced')
+
+    return (valueInDb && parseInt(valueInDb)) || this.earliestBlock
```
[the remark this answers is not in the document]
hasparus: I used `&&` `||` intentionally (maybe not very elegantly) to make sure this never fails, even if the kvStore contains a "corrupt value".

```
valueInDb = "database-is-corrupt?"

parseInt(valueInDb) // NaN
```

Also, when it's `0`, wouldn't we want to use `this.earliestBlock` instead?

packages/backend/test/peripherals/database/SyncStatusRepository.test.ts
sz-piotr: Then how the value 20 is being returned?

packages/backend/src/peripherals/database/SyncStatusRepository.ts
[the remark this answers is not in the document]
sz-piotr: But it can be corrupted and still return a number, e.g: `parseInt('3 is my favorite number')` returns `3`. If we want to check this case we should have a check like `/^\d+$/.test(valueInDb)`

packages/backend/test/peripherals/database/SyncStatusRepository.test.ts
hasparus: It could be hardcoded. We're just running one example.

packages/backend/src/peripherals/database/SyncStatusRepository.ts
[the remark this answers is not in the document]
hasparus: We could go with `Number` instead of `parseInt`

```
Number('3 is my favorite number') // NaN
```

packages/backend/test/peripherals/database/SyncStatusRepository.test.ts
sz-piotr: Ok let's leave it like this then

packages/backend/src/peripherals/database/SyncStatusRepository.ts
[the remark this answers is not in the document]
sz-piotr: Let's do `Number` then

sz-piotr (approved):

PR #21, opened 2022-01-12
Collect verifier events to database

hasparus:
**What's changed?**

- Added _VerifierCollector_ to store _ImplementationAdded_ and _Upgraded_ events in the database.
- Changed _SyncScheduler's_ `BATCH_SIZE` to 10000 — having 500 there caused very slow updates. How high can we get without causing problems? @sz-piotr 
- Added `003_verifier_events` db migration.

---

I've put all Upgraded and ImplementationAdded events in the same table for easier inspection.

image

Didn't use Postgres native enums, but we should be able to change this pretty easily with `table.enum('name', { enumName: 'VerifierEventName', useNative: true }`. I'm not sure if we want to.

### DIFF
--- a/packages/backend/src/Application.ts
+++ b/packages/backend/src/Application.ts
@@ -64,7 +64,7 @@ export class Application {
       safeBlockService,
       dataSyncService,
       logger,
-      config.core.sync.batchSize
+      config.core.syncBatchSize
     )
 
     // #endregion core
--- a/packages/backend/src/Application.ts
+++ b/packages/backend/src/Application.ts
@@ -63,7 +63,8 @@ export class Application {
       syncStatusRepository,
       safeBlockService,
       dataSyncService,
-      logger
+      logger,
+      config.core.sync.batchSize
     )
 
     // #endregion core
--- a/packages/backend/src/Application.ts
+++ b/packages/backend/src/Application.ts
@@ -7,10 +7,12 @@ import { DataSyncService } from './core/DataSyncService'
 import { SafeBlockService } from './core/SafeBlockService'
 import { StatusService } from './core/StatusService'
 import { SyncScheduler } from './core/SyncScheduler'
+import { VerifierCollector } from './core/VerifierCollector'
 import { DatabaseService } from './peripherals/database/DatabaseService'
 import { KeyValueStore } from './peripherals/database/KeyValueStore'
 import { PositionUpdateRepository } from './peripherals/database/PositionUpdateRepository'
 import { SyncStatusRepository } from './peripherals/database/SyncStatusRepository'
+import { VerifierEventRepository } from './peripherals/database/VerifierEventRepository'
 import { EthereumClient } from './peripherals/ethereum/EthereumClient'
 import { Logger } from './tools/Logger'
 
@@ -32,6 +34,8 @@ export class Application {
     const syncStatusRepository = new SyncStatusRepository(kvStore)
     // @todo unused for now
     new PositionUpdateRepository(knex, logger)
+    const verifierEventRepository = new VerifierEventRepository(knex, logger)
+
     const ethereumClient = new EthereumClient(config.jsonRpcUrl)
 
     // #endregion peripherals
@@ -49,7 +53,12 @@ export class Application {
       safeBlockService,
     })
 
-    const dataSyncService = new DataSyncService()
+    const verifierCollector = new VerifierCollector(
+      ethereumClient,
+      verifierEventRepository
+    )
+
+    const dataSyncService = new DataSyncService(verifierCollector, logger)
     const syncScheduler = new SyncScheduler(
       syncStatusRepository,
       safeBlockService,
@@ -74,6 +83,7 @@ export class Application {
       await databaseService.migrateToLatest()
 
       await apiServer.listen()
+      await safeBlockService.start()
       await syncScheduler.start()
 
       logger.for(this).info('Started')
--- a/packages/backend/src/core/VerifierCollector.ts
+++ b/packages/backend/src/core/VerifierCollector.ts
@@ -34,29 +34,18 @@ export class VerifierCollector {
    */
   async collect(blockRange: BlockRange): Promise<EthereumAddress[]> {
     const oldEvents = this.verifierEventRepository.getAll()
-    const newEvents = this.getEvents(blockRange)
-    const savingNewEventsToDb = newEvents.then((events) =>
-      this.verifierEventRepository.addOrUpdate(events)
-    )
+    const newEvents = await this.getEvents(blockRange)
+    const savingNewEventsToDb =
+      this.verifierEventRepository.addOrUpdate(newEvents)
 
-    const [oldUpgraded, oldAdded] = partitionVerifierEvents(await oldEvents)
-    const [newUpgraded, newAdded] = partitionVerifierEvents(await newEvents)
+    const events = [...(await oldEvents), ...newEvents]
+    const [upgraded, added] = partitionVerifierEvents(events)
 
-    oldAdded.sort(byBlockNumberDescending)
-    newAdded.sort(byBlockNumberDescending)
-
-    // verifiers we already knew about
-    const oldVerifiers = computeVerifiersFromEvents(oldAdded, oldUpgraded)
-
-    // verifiers upgraded in the block range
-    const newVerifiers = computeVerifiersFromEvents(
-      [...oldAdded, ...newAdded],
-      newUpgraded
-    )
+    added.sort(byBlockNumberDescending)
 
     await savingNewEventsToDb
 
-    return [...oldVerifiers, ...newVerifiers]
+    return computeVerifiersFromEvents(added, upgraded)
   }
 
   private async getEvents(blockRange: BlockRange) {
--- a/packages/backend/src/core/VerifierCollector.ts
+++ b/packages/backend/src/core/VerifierCollector.ts
@@ -35,8 +35,7 @@ export class VerifierCollector {
   async collect(blockRange: BlockRange): Promise<EthereumAddress[]> {
     const oldEvents = this.verifierEventRepository.getAll()
     const newEvents = await this.getEvents(blockRange)
-    const savingNewEventsToDb =
-      this.verifierEventRepository.addOrUpdate(newEvents)
+    const savingNewEventsToDb = this.verifierEventRepository.add(newEvents)
 
     const events = [...(await oldEvents), ...newEvents]
     const [upgraded, added] = partitionVerifierEvents(events)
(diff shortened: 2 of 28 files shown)


sz-piotr:
I'm hesitant to make the value higher than `6000` (a round number close to the number of blocks per day) since this would mean that the batches cover more than a single day at a time. I think it's fine for the initial sync to take some time, I'd be more concerned if we weren't able to keep up once we are synced.

hasparus:
> I'm hesitant to make the value higher than 6000 [...] I'd be more concerned if we weren't able to keep up once we are synced.

Makes sense. Actually, is the config a right place for this?

image

sz-piotr:
I think it's fine to have it in config, but you can use `getEnv.integer` instead of `parseInt`

packages/backend/src/peripherals/database/VerifierEventRepository.ts
```diff
@@ -0,0 +1,104 @@
...
+  name: 'ImplementationAdded'
+}
+
+export interface UpgradedEventRecord {
+  implementation: string
+  blockNumber: number
+  name: 'Upgraded'
+}
+
+export class VerifierEventRepository
```
sz-piotr, line 27: I would change the API surface of this class.
Basically what we need is the following:
- `add(records)`
- `getAll()`
- `getAllBefore(blockNumber)`
- `deleteAllAfter(blockNumber)`
- `deleteAll`
This makes it hard to justify implementing `Repository`, because it requires having `addOrUpdate` which we aren't going to use.

packages/backend/src/Application.ts
```diff
@@ -49,12 +53,18 @@ export class Application {
...
+    )
+
+    const dataSyncService = new DataSyncService(verifierCollector, logger)
     const syncScheduler = new SyncScheduler(
       syncStatusRepository,
       safeBlockService,
       dataSyncService,
-      logger
+      logger,
+      config.core.sync.batchSize
```
sz-piotr, line 67: I'm wondering if we need the nesting. Wouldn't just `config.core.batchSize` be enough?

packages/backend/src/peripherals/database/PositionUpdateRepository.ts
```diff
@@ -25,12 +25,16 @@ export class PositionUpdateRepository
   }
 
   async addOrUpdate(records: PositionUpdateRecord[]) {
+    if (records.length === 0)
+      return this.logger.debug({ method: 'addOrUpdate', rows: 0 })
```
sz-piotr, line 29: This couples the return type of logger.debug to this function. I'd prefer the longer option, when return is written on a new line.

packages/backend/src/peripherals/database/DatabaseService.ts
```diff
@@ -38,17 +38,19 @@ export class DatabaseService {
   }
 
   async migrateToLatest() {
+    const oldVersion = await this.knex.migrate.currentVersion()
     await this.knex.migrate.latest()
     const version = await this.knex.migrate.currentVersion()
     this.migrated = true
     this.version = version
-    this.logger.info('Migrations completed', { version })
+    this.logger.info('Migrations completed', { oldVersion, version })
```
sz-piotr, line 46: Nice :+1:

packages/backend/src/core/VerifierCollector.ts
```diff
@@ -0,0 +1,112 @@
...
+      this.verifierEventRepository.addOrUpdate(newEvents)
+
+    const events = [...(await oldEvents), ...newEvents]
+    const [upgraded, added] = partitionVerifierEvents(events)
+
+    added.sort(byBlockNumberDescending)
+
+    await savingNewEventsToDb
+
+    return computeVerifiersFromEvents(added, upgraded)
```
sz-piotr, line 48: This could be simplified. If `getEvents` also called awaited `verifierEventRepository.addOrUpdate` you could write:
```ts
const [oldEvents, newEvents] = await Promise.all([
  this.verifierEventRepository.getAll(),
  this.getEvents(blockRange)
])
```
And the code would be equivalent with the small exception that partitioning and computing happen after the events are persisted, which shouldn't make a noticeable difference

packages/backend/src/tools/Logger.ts
```diff
@@ -33,9 +33,9 @@ export class Logger {
     return this.configure({ service: object.constructor.name })
   }
 
-  error(error: unknown) {
+  error(error: unknown, annotation?: string) {
     if (this.options.logLevel >= LogLevel.ERROR) {
-      const message = getErrorMessage(error)
+      const message = [annotation, getErrorMessage(error)].join(' ')
```
sz-piotr, line 38: This will produce messages that have an extra `" "` at the start

packages/backend/src/tools/JobQueue.ts
```diff
@@ -57,7 +59,7 @@ export class JobQueue {
       await job.execute()
       this.inProgress.splice(this.inProgress.indexOf(job), 1)
     } catch (error) {
-      this.logger.error(error)
+      this.logger.error(error, `Job "${job.name}" failed with:`)
```
sz-piotr, line 62: Maybe we can detect `arguments.length` and write `this.logger.error(message, error)`

sz-piotr (changes requested):

packages/backend/src/Application.ts
hasparus: I'm not quite sure if `core.batchSize` is self-documenting enough, but I'll change it to `core.syncBatchSize`.

packages/backend/src/tools/Logger.ts
hasparus: Nice catch.

packages/backend/src/core/VerifierCollector.ts
hasparus: I'm a bit unsure about persisting events to database happening in method called `getEvents`, thus "collect".

If you feel that this late `await savingNewEventsToDb` is harming readability, I can await it where `add` is called.

packages/backend/src/peripherals/database/VerifierEventRepository.ts
hasparus: What do you feel about this:

```ts
interface Repository<TRecord> {
  addOrUpdate?(records: TRecord[]): Promise<void>
  add?(records: TRecord[]): Promise<void>
}
```

Translating to English, I'd read it as: _If a repository has a method called `add`, then it must take an array of `TRecord`._

packages/backend/src/peripherals/database/VerifierEventRepository.ts
```diff
@@ -47,13 +47,30 @@ export class VerifierEventRepository
   async getAll() {
     const rows = await this.knex('verifier_events').select('*')
     this.logger.debug({ method: 'getAll', rows: rows.length })
+
+    return rows.map(toRecord)
+  }
+
+  async getAllAfter(blockNumber: number) {
+    const rows = await this.knex('verifier_events')
+      .where('blockNumber', '>', blockNumber)
```
hasparus, line 56: Should this be inclusive or exclusive @sz-piotr ?

packages/backend/src/tools/Logger.ts
```diff
@@ -33,9 +33,15 @@ export class Logger {
     return this.configure({ service: object.constructor.name })
   }
 
-  error(error: unknown) {
+  error(error: unknown): void
+  error(annotation: string, error: unknown): void
+  error(...args: [unknown] | [string, unknown]): void {
+    const [annotation, error] = args.length === 1 ? ['', args[0]] : args
```
sz-piotr, line 39: I'd move this inside the if. There is no need to process any logic outside.

packages/backend/src/tools/JobQueue.ts
```diff
@@ -18,7 +18,9 @@ export class JobQueue {
   private inProgress: JobInQueue[] = []
   private lastUpdatedAt = new Date().toISOString()
 
-  constructor(private options: JobQueueOptions, private logger: Logger) {}
+  constructor(private options: JobQueueOptions, private logger: Logger) {
+    this.logger = logger.for(this)
```
sz-piotr, line 22: Not having this line here was on purpose. JobQueue is not an Application level sevice but a utility. Maybe logger can have a `nested` method, so that it formats as: `ServiceName.JobQueue` instead

packages/backend/src/peripherals/database/types.ts
```diff
@@ -19,14 +19,27 @@ declare module 'knex/types/tables' {
...
     position_updates: PositionUpdateRow
     key_values: KeyValueRow
+    verifier_events: VerifierEventRow
   }
 }
 
 export interface Repository<TRecord> {
-  addOrUpdate(records: TRecord[]): Promise<void>
+  addOrUpdate?(records: TRecord[]): Promise<void>
+  add?(records: TRecord[]): Promise<void>
```
sz-piotr, line 42: Ah, I like it now. It enforces a convention but isn't overwhelming.

packages/backend/src/core/VerifierCollector.ts
```diff
@@ -0,0 +1,111 @@
...
+    private readonly ethereumClient: Pick<EthereumClient, 'getLogs'>,
+    private readonly verifierEventRepository: VerifierEventRepository
+  ) {}
+
+  /**
+   * Saves new `Upgraded` and `ImplementationAdded` events to the database,
+   * and returns verifier addresses derived from both the old and new events.
+   */
+  async collect(blockRange: BlockRange): Promise<EthereumAddress[]> {
+    const oldEvents = this.verifierEventRepository.getAll()
```
sz-piotr, line 36: I though that we should only query events up to this point, so `.getAllBefore(blockRange)`, but because we are guaranteed to have no newer events then getAll may be fine.

packages/backend/src/peripherals/database/VerifierEventRepository.ts
sz-piotr: It was supposed to be `getAllBefore` and exclusive. But I'm not sure if we need this, see comment in VerifierCollector

packages/backend/src/tools/JobQueue.ts
hasparus: Hmm... shouldn't `for` always work like this?

If I have a `logger.for(syncScheduler).for(jobQueue)` formatting `ServiceName.JobQueue` seems like the expected result.

packages/backend/src/core/VerifierCollector.ts
hasparus: Should VerifierCollector have a `revert` or `cleanup` method then? We'd call it in DataSyncService `.revert` and remove all events after block number.

sz-piotr (approved):
Resolve conflicts with master and we can merge this

PR #23, opened 2022-01-13
Page events

hasparus:
description TBD


packages/backend/src/peripherals/database/PageRepository.ts
```diff
@@ -0,0 +1,93 @@
...
+    return rows.map(toRecord)
+  }
+
+  async getAllForFacts(factHashes: string[]) {
+    type Row = { fact_hash: string; pages: string[] }
+
+    const rows = (await this.knex('fact_to_pages')
+      .select(
+        'fact_hash',
+        this.knex.raw(
```
sz-piotr, line 44: Ah the beauty of raw SQL. I would have done it in JS unaware of the possibilities.

packages/backend/src/peripherals/database/migrations/004_fact_to_pages.ts
```diff
@@ -0,0 +1,28 @@
...
+should create a new migration file that fixes the issue.
+
+*/
+
+import { Knex } from 'knex'
+
+export async function up(knex: Knex) {
+  await knex.schema.createTable('fact_to_pages', (table) => {
+    table.increments('id').primary()
+    table.integer('block_number').notNullable()
```
sz-piotr, line 19: At some point we'll need to set up database indices on columns like `block_number`, but that can be done in the future

packages/backend/src/core/PageCollector.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  async collect(blockRange: BlockRange): Promise<PageRecord[]> {
+    const memoryPageEvents = await this.getMemoryPageEvents(blockRange)
+
+    const records = await Promise.all(
+      memoryPageEvents.map(
+        async ({ memoryHash, transactionHash }): Promise<PageRecord> => {
+          const tx = await this.ethereumClient.getTransaction(transactionHash)
+
+          const decoded = PAGE_ABI.decodeFunctionData(
```
sz-piotr, line 39: This isn't tested. Instead of having a huge blob of data in tests i suggest using `PAGE_API.encodeFunctionData` and checking that the values are actually what is being decoded as a page.

sz-piotr (changes requested):

packages/backend/src/core/DataSyncService.ts
```diff
@@ -1,27 +1,72 @@
...
+    private readonly memoryHashEventCollector: MemoryHashEventCollector,
+    private readonly pageCollector: PageCollector,
+    private readonly stateTransitionFactCollector: StateTransitionFactCollector,
+    // @todo where to move this?
+    private readonly pageRepository: Pick<PageRepository, 'getAllForFacts'>,
+    private readonly positionUpdateRepository: Pick<
+      PositionUpdateRepository,
+      'addOrUpdate'
+    >,
+    // @todo ^ and this, should it be here?
```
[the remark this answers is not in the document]
hasparus: To be fair, I wrote this without thinking too much about it, but decided to leave it as a topic for conversation. `DataSyncService` only creates records in PositionUpdateRepository, and here in the constructor we "promise that".

I'll probably remove this Pick to simplify the code when removing todo comments, but I wanted to ask what do you think about it.

packages/backend/src/peripherals/database/PageRepository.ts
hasparus: I keep thinking about putting this in a string array from the get go instead of splitting it on insertion and aggregating it back on read.

packages/backend/src/core/DataSyncService.ts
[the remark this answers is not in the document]
sz-piotr: I think that while it has some small benefits the code complexity outweighs those benefits

PR #24, opened 2022-01-14
Move from chai to earljs

hasparus:


packages/backend/test/mock.ts
```diff
@@ -1,18 +1,31 @@
...
-  const clone = { ...overrides }
+import { Mock as MockFunction, mockFn } from 'earljs'
+
+export type MockedObject<T> = {
+  [P in keyof T]: T[P] extends (...args: any[]) => any
+    ? MockFunction.Of<T[P]>
+    : T[P]
+} & T
+
+export function mock<T>(overrides: Partial<T> = {}): MockedObject<T> {
```
hasparus, line 9: Let me know what you think about this.

This function will now turn `{}` overrides into a mock object where every method is a mock function made with Earl's `mockFn`.

packages/backend/test/mock.ts
```diff
@@ -1,18 +1,31 @@
...
-      if (value !== undefined) {
+      if (value && typeof value !== 'function') {
         return value
       }
-      const name = String(property)
-      return () => {
-        throw new Error(
-          `Cannot call .${name}() - no mock implementation provided.`
-        )
+      if (isMockFunction(value)) {
```
sz-piotr, line 17: If we did the transformation before creating the proxy we could keep the old logic inside the proxy.

packages/backend/test/mock.ts
```diff
@@ -1,18 +1,31 @@
...
+        return value
       }
+      const m = mockFn(value || (() => {}))
```
sz-piotr, line 20: `mock<Foo>({ x: 0 })` would create a mock with x being a `mockFn`

packages/backend/test/mock.ts
```diff
@@ -1,18 +1,31 @@
...
+      Reflect.set(target, property, m, receiver)
+      return m
     },
   })
-  return proxy as T
+
+  return proxy
+}
+
+function isMockFunction(x: unknown): x is MockFunction<any[], unknown> {
```
sz-piotr, line 29: Ideally this could be exported from earl

packages/backend/test/core/StatusService.test.ts
```diff
@@ -9,7 +9,7 @@ describe('StatusService', () => {
       barService: new BarService(),
     })
 
-    expect(statusService.getStatus()).to.deep.eq({
+    expect<object>(statusService.getStatus()).toEqual({
```
sz-piotr, line 12: Use either `unknown` or `object` everywhere. I am not sure how i like the fact that this fails for infinite types.

sz-piotr:
Let me know what you think of the changes to `mock` and we can merge

PR #25, opened 2022-01-18
Test database fixes

hasparus:
There are two changes on this branch

1. I've set up a new schema for each test suite connecting to the database.
  - It solves the problem of switching between branches with different migrations, as we drop the schema in _before_ step.
2. Fixed a test failing on master. That's a bit troubling, as I just removed `Promise.all`. Is Knex giving us resolved promises before the data is written to the database?

### DIFF
--- a/packages/backend/test/peripherals/database/KeyValueStore.test.ts
+++ b/packages/backend/test/peripherals/database/KeyValueStore.test.ts
@@ -16,21 +16,22 @@ describe(KeyValueStore.name, () => {
     kvStore.set('key', 'value')
     const actual = await kvStore.get('key')
     expect(actual).toEqual('value')
+    await kvStore.delete('key')
   })
 
   it('reads and removes all values', async () => {
-    await Promise.all([
-      kvStore.set('1', 'one'),
-      kvStore.set('2', 'two'),
-      kvStore.set('3', 'three'),
-    ])
+    await kvStore.set('1', 'one')
+    await kvStore.set('2', 'two')
+    await kvStore.set('3', 'three')
 
     let actual = await kvStore.getAll()
-    expect(actual).toEqual([
-      { key: '1', value: 'one' },
-      { key: '2', value: 'two' },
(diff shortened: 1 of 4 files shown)


packages/backend/test/peripherals/database/KeyValueStore.test.ts
```diff
@@ -16,21 +16,22 @@ describe(KeyValueStore.name, () => {
     kvStore.set('key', 'value')
     const actual = await kvStore.get('key')
     expect(actual).toEqual('value')
+    await kvStore.delete('key')
   })
 
   it('reads and removes all values', async () => {
-    await Promise.all([
```
sz-piotr, line 22: I don't understand. As long as you await knex's promise in the repository it should be fine. Why would Promise.all not work?

packages/backend/test/peripherals/database/setup.ts
```diff
@@ -6,12 +8,27 @@ export function setupDatabaseTestSuite() {
...
+      // 'test_one_two' schema before running tests.
+
+      const titlePath = this.test?.parent?.titlePath()
+      schemaName = `test_${titlePath?.join('_') || uuid()}`
+
+      log('Creating test schema', schemaName)
+
+      // We drop before instead of after tests, so we can inspect the
+      // contents of database when tests fail.
+      await knex.raw(`DROP SCHEMA IF EXISTS "${schemaName}" CASCADE`)
```
sz-piotr, line 27: You should use knex raw's parameters not string interpolation. This way no matter what the describe text is the query won't break

sz-piotr (changes requested):

packages/backend/test/peripherals/database/setup.ts
hasparus: I changed `DROP` and `CREATE` to use Knex parameter binding, but I'm unable to write `SET SCHEMA` like this :/

PR #33, opened 2022-01-20
Add nyc to track test coverage

hasparus:


packages/backend/package.json
```diff
@@ -12,6 +12,7 @@
     "typecheck": "tsc --noEmit",
     "build": "tsc",
     "test": "mocha",
+    "coverage": "nyc --reporter=lcov --reporter=text mocha",
```
sz-piotr, line 15: In the future we might want to add such scripts to all packages and have a repository level script too.

sz-piotr (approved):

PR #34, opened 2022-01-20
Add config.freshStart

hasparus:


packages/backend/src/config/config.production.ts
```diff
@@ -22,5 +22,6 @@ export function getProductionConfig(): Config {
       },
       syncBatchSize: getEnv.integer('SYNC_BATCH_SIZE', 6_000),
     },
+    freshStart: getEnv.boolean('FRESH_START', false),
```
sz-piotr, line 25: I don't think we should make dropping the production database that easy :) I suggest keeping it false

sz-piotr (changes requested):

PR #35, opened 2022-01-21
Improve MerkleTree

sz-piotr:


packages/state/test/MerkleTree.test.ts
```diff
@@ -261,29 +261,14 @@ describe(MerkleTree.name, () => {
...
+      tree = await tree.update([
         { index: 2n, value: positionA },
         { index: 7n, value: positionB },
       ])
 
       const recovered = new MerkleTree(storage, 3n, await tree.hash())
       expect(await recovered.getLeaf(2n)).toEqual(positionA)
     })
-
-    it('set hash changes the tree', async () => {
```
hasparus, line 229: If immutability is a feature we'd like to support, what do you think about adding a test case for it?

Something like

> it('does not mutate itself on update')

sz-piotr: Good idea

hasparus (approved):
LGTM

PR #40, opened 2022-01-25
Handle reorganizations

hasparus:


packages/backend/src/core/BlockDownloader.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  getLastKnownBlock() {
+    if (this.state.t !== 'working') {
+      throw new Error('Not started')
+    }
+    return this.state.lastKnownBlock
+  }
+
+  onNewBlocks(handler: (blockRange: BlockRange) => void) {
+    this.events.on('newBlocks', handler)
```
sz-piotr, line 62: What about this being called immediately with some value if we know of a last block?

sz-piotr (changes requested):

packages/backend/src/core/BlockDownloader.ts
hasparus: I don't think this is a correct behaviour.

If we call the listener with _just the last block_, we're neither _hot_ nor _cold_.

Current implementation is _hot_ like an usual event emitter. The events are emitted regardless if somebody listens or not. 

Since we're now reacting on `block` events instead of scheduling work on interval, I moved to handling "sync on restart" behaviour in SyncScheduler. I still have to build the "empowered BlockRange" here, but you can notice I started emitting `{from, to}` instead of just `to: number`.

Take a look at the following.

```ts
    if (lastKnown.number > lastSynced) {
      this.scheduleSync({ from: lastSynced, to: lastKnown.number })
      lastSynced = lastKnown.number
    }

    this.blockDownloader.onNewBlocks((blockRange) => {
      lastSynced = blockRange.to
      this.scheduleSync(blockRange)
    })
```

packages/backend/src/peripherals/database/types.ts
```diff
@@ -52,13 +52,20 @@ declare module 'knex/types/tables' {
     hash: string
   }
 
+  interface BlockRow {
+    /** surrogate key */
```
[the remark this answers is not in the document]
hasparus: Nice catch

packages/backend/src/core/BlockDownloader.ts
```diff
@@ -0,0 +1,218 @@
...
+        this.ethereumClient.getBlock(blockNumber)
+      )
+    )
+    newBlocks.push(latest)
+
+    // We fetch and save all blocks from the last known block, validating their
+    // parentHashes to ensure they form a valid chain, if they don't, we go back
+    // to the beginning of the procedure.
+    if (!isConsistentChain(newBlocks)) {
+      if (retriesRemaining > 0)
```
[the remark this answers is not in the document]
hasparus: Oh yeah, I wanted to talk about it.

So, currently jobQueue restarts all failed jobs.
- As inconsistent incoming blocks are something we expect, I don't want to throw an error on it — we expect it can happen, and we'll just try again.
- However, if it's not consistent after a few tries, I think we'd like to abort here, log an error that will land in an error reporting software and just forget about this job.

packages/backend/src/Application.ts
```diff
@@ -46,16 +47,16 @@ export class Application {
...
 
     const ethereumClient = new EthereumClient(config.jsonRpcUrl)
 
     // #endregion peripherals
     // #region core
 
-    const safeBlockService = new SafeBlockService(
-      config.core.safeBlock.refreshIntervalMs,
-      config.core.safeBlock.blockOffset,
+    const safeBlockService = new BlockDownloader(
```
sz-piotr, line 57: Rename this variable to blockDownloader

packages/backend/src/core/BlockDownloader.ts
sz-piotr: Makes sense. Let's continue that way

packages/backend/src/core/SyncScheduler.ts
```diff
@@ -23,31 +23,38 @@ export class SyncScheduler {
...
   async start() {
-    let last = await this.statusRepository.getLastBlockNumberSynced()
+    let lastSynced = await this.statusRepository.getLastBlockNumberSynced()
+    const lastKnown = this.blockDownloader.getLastKnownBlock()
 
-    this.logger.info('Last block number synced', { last })
+    this.logger.info('start', { lastSynced, lastKnown: lastKnown.number })
 
-    await this.dataSyncService.revert(last)
+    await this.dataSyncService.discard({ from: lastSynced })
```
sz-piotr, line 31: I think that replacing `discard({ from })` with `discardAfter(block)` improves the readability.

sz-piotr (changes requested):

packages/backend/src/core/BlockDownloader.ts
sz-piotr: Makes sense. Let's continue that way

packages/backend/src/core/BlockDownloader.ts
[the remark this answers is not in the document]
sz-piotr: I like the idea of a retry: false flag. However, what would be even more awesome is a maxRetries, then you can set it to 2 and not have any recursion.

packages/backend/src/core/BlockDownloader.ts
hasparus: Alternatively, we coould make it entirely _cold_ and replay all unhandled events from the start of the downloader for every new listener. It would be "cleaner but trickier", so I'm not convinced if we wanna go in this direction.

packages/backend/src/core/SyncScheduler.ts
hasparus: I went with `discardFrom` to avoid changing all places wih `- 1`. LMK if that's ok.

packages/backend/src/model/BlockRange.ts
```diff
@@ -0,0 +1,76 @@
...
+  public readonly to: BlockNumber
+  private readonly hashes: ReadonlyMap<BlockNumber, string>
+
+  constructor(
+    /** a list of blocks in order from oldest to newest */
+    blocks: readonly Block[]
+  )
+  constructor(
+    /** parent block range */
+    range: BlockRange,
```
sz-piotr, line 19: This should probably be either a range or a list of blocks, so that one doesn't have to write `new BlockRange(new BlockRange(blocks), from, to)` if the start and end have unknown hashes

packages/backend/src/model/BlockRange.ts
```diff
@@ -0,0 +1,76 @@
...
+      this.to = to
+      this.from = from
+      this.hashes = range.hashes
+      return
+    }
+
+    const [blocks] = args
+    assert(blocks.length > 0, 'BlockRange must have at least one block')
+
+    this.from = blocks[0].number
```
sz-piotr, line 38: This should be the smallest, we aren't guaranteed that the blocks are given in-order

packages/backend/src/model/BlockRange.ts
```diff
@@ -0,0 +1,76 @@
...
+    this.to = blocks[blocks.length - 1].number
```
sz-piotr, line 39: Similarily this should be largest

packages/backend/src/model/BlockRange.ts
```diff
@@ -0,0 +1,76 @@
...
+      blocks.map((block) => [block.number, block.hash] as const)
+    )
+  }
+
+  has({
+    blockNumber,
+    blockHash,
+  }: {
+    blockNumber: BlockNumber
+    blockHash: string
```
sz-piotr, line 50: Hash256

packages/backend/src/model/BlockRange.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  static fake({ from, to }: { from: BlockNumber; to: BlockNumber }) {
+    return new BlockRange(
+      range(from, to + 1).map((i) => ({ number: i, hash: '0x' + i }))
+    )
+  }
+}
+
+type Block = { number: BlockNumber; hash: string }
```
sz-piotr, line 73: hash: Hash256

sz-piotr: Also, this should be exported and potentially renamed to BlockInRange

packages/backend/src/peripherals/database/BlockRepository.ts
```diff
@@ -0,0 +1,107 @@
...
+ * @internal exported for tests
+ */
+export const EARLIEST_BLOCK: Readonly<BlockRecord> = {
+  number: 11813207,
+  hash: '0xe191f743db9d988ff2dbeda3ec800954445f61cf8e79cc458ba831965e628e8d',
+}
+
+export interface BlockRecord {
+  number: number
+  hash: string
```
sz-piotr, line 18: Hash256

packages/backend/src/core/BlockDownloader.ts
```diff
@@ -0,0 +1,239 @@
...
+            this.logger.error(error)
+          }
+        },
+      })
+    )
+  }
+
+  private async handleNewBlock(
+    latest: IncomingBlock,
+    retriesRemaining = 2
```
sz-piotr, line 94: You can merge #47 and then remove recursion

packages/backend/src/core/SyncScheduler.ts
```diff
@@ -23,50 +26,64 @@ export class SyncScheduler {
   )
 
   async start() {
-    let last = await this.statusRepository.getLastBlockNumberSynced()
+    let lastSynced = await this.statusRepository.getLastBlockNumberSynced()
+    await this.dataSyncService.discardFrom(lastSynced)
```
sz-piotr, line 30: Does it make sense now? This means that we are removing the last synced block.

packages/backend/src/core/BlockDownloader.ts
```diff
@@ -0,0 +1,218 @@
...
+import { createEventEmitter } from '../tools/EventEmitter'
+import { JobQueue } from '../tools/JobQueue'
+import { Logger } from '../tools/Logger'
+
+export interface KnownBlock {
+  number: number
+  hash: string
+}
+
+type State = { t: 'not-started' } | { t: 'working'; lastKnownBlock: KnownBlock }
```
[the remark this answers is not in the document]
sz-piotr: Bump

packages/backend/src/core/PageCollector.ts
```diff
@@ -74,6 +74,7 @@ export class PageCollector {
     })
 
     return logs
+      .filter((log) => blockRange.has(log))
```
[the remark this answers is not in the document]
sz-piotr: This is applicable to all the places that fetch logs

packages/backend/src/core/SyncScheduler.ts
```diff
@@ -23,50 +26,64 @@ export class SyncScheduler {
...
+    if (lastKnown.number > lastSynced) {
+      const blocks = await this.blockRepository.getAllInRange(
+        lastSynced + 1,
+        lastKnown.number
+      )
+      lastSynced = lastKnown.number
+      this.scheduleSync(new BlockRange(blocks))
+    }
+
+    this.blockDownloader.onNewBlocks(async (newBlockRange) => {
```
sz-piotr, line 45: Isn't there a race condition here that between getting the last known block and subscribing the last known block actually changes and we miss some blocks?

sz-piotr (approved):

packages/backend/test/core/SyncScheduler.test.ts
```diff
@@ -1,143 +1,638 @@
...
+        expect(newState).toEqual({
+          ...initialState,
+          discardRequired: true,
+          blocksToProcess: new ContinuousBlocks(newBlocks),
+          latestBlockProcessed: 3,
+        })
+
+        // There is no `discard` yet, because we need to process old block 4 still.
+        expect(effects).toEqual([])
+        ;[newState, effects] = syncSchedulerReducer(newState, {
```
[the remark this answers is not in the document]
hasparus: I could bet you destructured function parameters (e.g. React props) at least once.

PR #41, opened 2022-01-27
Fix page hash padding

sz-piotr:


packages/backend/src/core/PageCollector.ts
```diff
@@ -74,12 +74,13 @@ export class PageCollector {
 
     return logs
       .map((log) => ({ log, event: REGISTRY_ABI.parseLog(log) }))
-      .map(
-        ({ log, event }): MemoryPageEvent => ({
-          memoryHash: event.args.memoryHash.toHexString(),
+      .map(({ log, event }): MemoryPageEvent => {
+        const hash: BigNumber = event.args.memoryHash
+        return {
+          memoryHash: '0x' + hash.toHexString().slice(2).padStart(64, '0'),
```
hasparus, line 80: Oh dude, this looks very familiar :D

We have the following function used in `collect` in this exact file.

```ts
/** @internal */
export function bignumToPaddedString(x: BigNumber): string {
  return x.toHexString().substring(2).padStart(64, '0')
}
```


https://github.com/l2beat/dydx-state-explorer/blob/cb8186eae57f79ffd110c1b75f61df9cc6995434/packages/backend/src/core/PageCollector.ts#L93

---

I'll remove it and turn its uses to `Hash256(x).toString())`.

PR #43, opened 2022-01-28
Add Hash256

sz-piotr:
Requires #42

### DIFF
--- a/packages/backend/test/model/Hash256.test.ts
+++ b/packages/backend/test/model/Hash256.test.ts
@@ -1,0 +1,65 @@
+import { expect } from 'earljs'
+import { BigNumber } from 'ethers'
+
+import { Hash256 } from '../../src/model'
+
+describe(Hash256.name, () => {
+  it('accepts a valid hash', () => {
+    const hash = Hash256(
+      '0xabcd1234abcd1234abcd1234abcd1234abcd1234abcd1234abcd1234abcd1234'
+    )
+    expect(hash.toString()).toEqual(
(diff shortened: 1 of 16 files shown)


packages/backend/test/model/Hash256.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  it('throws for short hex strings', () => {
+    expect(() => Hash256('0x123abc')).toThrow(TypeError)
+  })
+
+  it('throws for long hex strings', () => {
+    expect(() => Hash256('0x' + '1'.repeat(65))).toThrow(TypeError)
+  })
+
+  describe(Hash256.fromBigInt.name, () => {
```
hasparus, line 46: I'm not very opinionated on this but wouldn't it be more idiomatic in TS/JS to have just one constructor function that takes a union of inputs?

packages/backend/test/peripherals/database/FactToPageRepository.test.ts
```diff
@@ -17,8 +18,8 @@ describe(FactToPageRepository.name, () => {
     const record: FactToPageRecord = {
       blockNumber: 1,
       index: 0,
-      factHash: '{{ fact hash }}',
-      pageHash: '{{ page hash }}',
+      factHash: Hash256.fake(),
```
hasparus, line 21: I love it.

packages/backend/test/model/Hash256.test.ts
sz-piotr: People have different views on this but I'm happy to refactor.

hasparus (approved):

PR #53, opened 2022-02-04
Save decoded state to database

hasparus:


packages/backend/src/peripherals/database/types.ts
```diff
@@ -52,13 +39,57 @@ declare module 'knex/types/tables' {
     hash: string
   }
 
+  interface BlockRow {
+    number: number
+    hash: string
+  }
+
+  interface StateUpdateRow {
+    id: number
```
hasparus, line 48:
```suggestion
    id: number
    block_number: number
```

hasparus:
🎉 It still needs some more tests, but I guess it's good for review.

packages/state/src/RollupState.ts
```diff
@@ -102,7 +102,10 @@ export class RollupState {
       funding,
     })
 
-    return new RollupState(this.storage, positions, timestamp, funding)
+    return [
+      new RollupState(this.storage, positions, timestamp, funding),
+      newPositions,
```
sz-piotr, line 107: It's missing at least one small test for this

packages/backend/src/core/DataSyncService.ts
```diff
@@ -46,10 +62,42 @@ export class DataSyncService {
       stateTransitionFacts.map((f) => f.hash)
     )
 
-    for (const { pages } of stateTransitions) {
+    let { oldHash, id } = await this.readLastUpdate()
```
sz-piotr, line 65: Maybe this part of the `sync` method (from this line downwards) could be extracted into it's own class? It introduces 3 new dependencies to `DataSyncService` and is also a bit hard to test. If we instead extract it it would be a lot easier to manage.

sz-piotr (changes requested):

packages/backend/test/peripherals/database/StateUpdateRepository.test.ts
```diff
@@ -0,0 +1,228 @@
...
+          timestamp: 0,
+        },
+        positions: [],
+        prices: [],
+      })
+    }
+
+    await repository.deleteAllAfter(20_002)
+
+    const records = await repository.getAll()
```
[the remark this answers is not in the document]
hasparus: Yeah, I actually did it for `.delete()`, but we don't really need `.delete()` so I'll just move that part of the test here

https://github.com/l2beat/dydx-state-explorer/pull/53/files/e6f97e1df3adbe62b7392b886c1d2687b98f8a74#diff-5b80ae7736b12ab0a274f98b8351ed89a9e8ab25ba430cb409b5330fdc1275e0R90-R98

packages/backend/src/core/DataSyncService.ts
hasparus: Good idea!

packages/backend/src/peripherals/database/StateUpdateRepository.ts
```diff
@@ -0,0 +1,166 @@
...
+}
+
+export class StateUpdateRepository {
+  constructor(private knex: Knex, private logger: Logger) {
+    this.logger = logger.for(this)
+  }
+
+  async add({ stateUpdate, positions, prices }: StateUpdateBundle) {
+    await this.knex.transaction(async (trx) => {
+      await this.knex('state_updates')
```
sz-piotr, line 36: I think that writing `trx('...')` is cleaner than `this.knex('...').transacting(trx)`, but if you feel differently we can keep the longer version.

packages/backend/src/core/StateUpdateCollector.ts
```diff
@@ -0,0 +1,112 @@
...
+        stateUpdate: {
+          id: ++id,
+          blockNumber,
+          factHash,
+          rootHash,
+          timestamp,
+        },
+        positions: newPositions.map(
+          ({ value, index }): PositionRecord => ({
+            positionId: Number(index), // @todo This isn't correct, right? How do I get `positionId`?
```
sz-piotr, line 65: I think this is actually correct, but I'm not sure if we can cast it to Number, because there are up to 2**64 positions.

sz-piotr (changes requested):

PR #58, opened 2022-02-09
Improve BlockRange

sz-piotr:
I need this to fix SyncScheduler

### DIFF
--- a/packages/backend/src/peripherals/ethereum/EthereumClient.ts
+++ b/packages/backend/src/peripherals/ethereum/EthereumClient.ts
@@ -1,6 +1,7 @@
+import { assert } from 'console'
 import { ethers, providers } from 'ethers'
 
-import { Hash256 } from '../../model'
+import { BlockRange, Hash256 } from '../../model'
 import { BlockTag } from './types'
 
 export class EthereumClient {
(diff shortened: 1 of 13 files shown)


packages/backend/src/peripherals/ethereum/EthereumClient.ts
```diff
@@ -20,6 +21,19 @@ export class EthereumClient {
...
+      return []
+    }
+    const logs = await this.provider.getLogs({
+      ...filter,
+      fromBlock: blockRange.start,
+      toBlock: blockRange.end - 1,
+    })
+    assert(blockRange.hasAll(logs), 'all logs must be from the block range')
+    return logs
+  }
```
hasparus, line 35: I like this change a lot, but I'm curious about the rename from `from` and `to` to `start` and `end`? Does `end` communicate exclusivity better?

hasparus: > It would be just too convinient to pass from and to into Filter's fromBlock and toBlock, so I changed the naming so that maybe some warning light would flash in the programmer's mind

👍 Makes total sense.

hasparus (approved):

PR #68, opened 2022-02-22
Extract routers logic into controllers

sin6pi7:


packages/backend/src/api/routers/FrontendRouter.ts
```diff
@@ -1,95 +1,33 @@
...
-        timestamp: update.timestamp,
-        positionCount: update.positionCount,
-      })),
-      fullCount: Number(fullCount),
-      params: {
-        page,
-        perPage,
-      },
-    })
+    if ([page, perPage].some(Number.isNaN)) {
```
sz-piotr, line 16: We'd also want to change this logic so that we parse `ctx.query` using zod. This can be implemented as part of another PR.

sz-piotr (changes requested):

packages/backend/src/api/routers/FrontendRouter.ts
sin6pi7: These were my thoughts as well. However, the router which is responsible for request data validation and parsing was not doing that for this specific route (unlike other routes where `PedersenHash` and `BigInt` constructors were handling it), so I added it to stay consistent. Otherwise we rely on the controller.

PR #69, opened 2022-02-23
Unify request validation

sin6pi7:
To figure out:
1. Typing of `withRequestDataParsing` to enforce input shape to have `body`, `params` and `query` keys,
2. Where to put helpers.

### DIFF
--- a/packages/backend/src/api/routers/FrontendRouter.ts
+++ b/packages/backend/src/api/routers/FrontendRouter.ts
@@ -2,31 +2,8 @@ import { PedersenHash } from '@explorer/crypto'
 import Router from '@koa/router'
 import { z } from 'zod'
 
-import {
-  brandedString,
-  stringToPositiveInt,
-  withRequestDataParsing,
-} from '../../tools/RequestDataParsing'
 import { FrontendController } from '../controllers/FrontendController'
-
-const GetStateChangesParser = z.object({
-  query: z.object({
-    page: stringToPositiveInt('1'),
-    perPage: stringToPositiveInt('10'),
-  }),
-})
-
-const GetStateChangeDetailsParser = z.object({
-  params: z.object({
-    hash: brandedString(PedersenHash),
-  }),
-})
-
-const GetPositionDetailsParser = z.object({
-  params: z.object({
-    positionId: brandedString(BigInt),
-  }),
-})
+import { brandedString, stringToPositiveInt, withTypedContext } from './types'
 
 export function createFrontendRouter(frontendController: FrontendController) {
   const router = new Router()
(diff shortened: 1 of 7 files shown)


packages/backend/src/tools/RequestDataParsing.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+export function withRequestDataParsing<T extends z.AnyZodObject>(
+  parser: T,
+  handler: (
+    ctx: Application.ParameterizedContext,
+    data: z.infer<T>
+  ) => Promise<void>
+) {
+  return async (_ctx: Application.ParameterizedContext) => {
```
sz-piotr: Is there a reason why this is named `_ctx` and not `ctx`?

packages/backend/src/tools/RequestDataParsing.ts
```diff
@@ -0,0 +1,45 @@
...
+    const parseResult = parser.safeParse({
+      body: _ctx.body,
+      params: _ctx.params,
+      query: _ctx.query,
+    })
+    if (!parseResult.success) {
+      _ctx.status = 400
+      _ctx.body = { issues: parseResult.error.issues }
+      return
+    }
```
sz-piotr: I like this for now, but in the future we might want to make `withRequestDataParsing` take two callbacks, similarily to `Promise.then(onSuccess, onError)`, so that we can return custom error pages.

packages/backend/src/tools/RequestDataParsing.ts
```diff
@@ -0,0 +1,45 @@
+import Application from 'koa'
+import { z } from 'zod'
+
+export function stringToPositiveInt(def: string) {
+  return z.preprocess(
+    (s) => parseInt(z.string().parse(s || def), 10),
```
sz-piotr: Doesn't this parse `"123abc"` as `123`?

packages/backend/src/api/routers/FrontendRouter.ts
```diff
@@ -10,25 +16,51 @@ export function createFrontendRouter(frontendController: FrontendController) {
...
+          page: stringToPositiveInt('1'),
+          perPage: stringToPositiveInt('10'),
+        }),
+      }),
+      async (ctx) => {
+        const { page, perPage } = ctx.query
+        ctx.body = await frontendController.getStateChangesPage(page, perPage)
+      }
+    )
+  )
```
sz-piotr, line 33: I really liked your solution, I just had a thought on how can this actually be improved. I wasn't sure if changing the type of `ctx` can actually work, so I quickly changed the code to see if it was possible, but once I had the code I just decided that I can push it instead of making you implement the change again :)

sz-piotr: Let me know what do you think about this change.

sz-piotr (changes requested):

sz-piotr (approved):

PR #75, opened 2022-03-15
Web UI

sin6pi7:


packages/frontend/src/pages/common/Page.tsx
```diff
@@ -1,20 +1,37 @@
...
   scripts: string[]
   children: ReactNode
+  navbarSearch?: boolean
 }
 
-export function Page({ children, scripts, ...head }: Props) {
+export function Page({
+  children,
+  scripts,
+  navbarSearch = true,
```
sz-piotr, line 15: I don't like when optional parameters default to something other than undefined, because it introcuded ambiguity. Consider the following:
```tsx
<Page />
<Page navbarSearch />
```
Instead you could have a parameter like `withoutSearch` that defaults to undefined.

packages/frontend/src/pages/common/Pagination.tsx
```diff
@@ -0,0 +1,124 @@
...
+import cx from 'classnames'
+
+type PaginationProps = {
+  page: number
+  perPage: number
+  fullCount: number
+  baseUrl?: string
+}
+
+const PrevIcon = (props: React.SVGProps<SVGSVGElement>) => (
```
sz-piotr, line 11: Can we move the icons to common/icons?

packages/frontend/src/pages/common/Table.tsx
```diff
@@ -0,0 +1,98 @@
...
+  rows: Row[]
+  className?: string
+}
+
+const textAlign = (numeric?: boolean) => (numeric ? 'text-right' : 'text-left')
+const cellPaddings = 'px-2 py-0.5'
+const cellOverflowStyles = 'text-ellipsis overflow-hidden'
+const maxWidthStyles = (maxWidth?: boolean) => maxWidth && 'max-w-[320px]'
+
+export function Table({ columns, rows, className }: TableProps) {
```
sz-piotr, line 29: I like how configurable this is

sz-piotr (changes requested):
Good work overall, left some small comments. LGTM

packages/frontend/src/pages/common/Page.tsx
sin6pi7: In general I agree, but in this case we only have one Page that does not need a navbar, which results in less code overall.

packages/frontend/src/pages/common/Pagination.tsx
sz-piotr: Some of it, but not those ones in pagination

sz-piotr (approved):

PR #77, opened 2022-03-17
Homepage stats

sin6pi7:


packages/frontend/src/scripts/index.ts
```diff
@@ -1,2 +1,18 @@
+import { formatLargeNumber } from '../pages/formatLargeNumber'
+import { tvlElId } from '../pages/home/Home'
```
sz-piotr, line 2: Can you make sure that this is tree shaken so that only the id exists in the final code?

packages/backend/src/peripherals/database/StateUpdateRepository.ts
```diff
@@ -182,6 +182,16 @@ export class StateUpdateRepository {
...
     this.logger.debug({ method: 'deleteAllAfter', rows: rowsCount })
   }
+
+  async countStateUpdates() {
+    const [{ count }] = await this.knex('state_updates').count({ count: '*' })
+    return count ? BigInt(count) : 0n
+  }
+
+  async countPositions() {
+    const [{ count }] = await this.knex('positions').count({ count: '*' })
```
sz-piotr, line 192: Is this correct? It should count unique ids from this table

sz-piotr (changes requested):

packages/frontend/src/scripts/index.ts
sin6pi7: Done

sz-piotr (approved):

PR #84, opened 2022-03-29
Add decoding for the full on chain data

sz-piotr:
The decoding for forced operations is still experimental and I expect it to break. Further analysis has to be done, to find a real world example and test on it, as the one we're using has 0 forced operations.


sz-piotr:
@sin6pi7 The code now works with real data, care to review :) ?

packages/encoding/src/data/index.ts
```diff
@@ -6,9 +6,10 @@ run()
 async function run() {
   const data = await getOnChainData(
     '0x46c212912be05a090a9300cf87fd9434b8e8bbca15878d070ba83375a5dbaebd'
+    // '0x5a81a54eb926a240264f9a481810242987ae566e82f29bcdd8cd32063e1976c5'
```
sin6pi7, line 9: Why is it commented out?

packages/encoding/test/readConditions.test.ts
```diff
@@ -0,0 +1,38 @@
...
+import { readToDecode } from './readToDecode'
+
+describe(readConditions.name, () => {
+  const decode = readToDecode(readConditions)
+
+  it('fails for empty data', () => {
+    expect(() => decode('')).toThrow(DecodingError, 'Went out of bounds')
+  })
+
+  it('can read zero conditions', () => {
```
sin6pi7, line 15: IMO `can` word does not bring any value in test case title - if we made it `reads zero conditions` it says the same thing, but is shorter at the same time. Wdyt?

sin6pi7 (approved):
Only cosmetic changes requested - overall LGTM 👍

packages/encoding/src/data/index.ts
sz-piotr: The first hash allows you to get the original data we had, the second is for getting data with forced actions

packages/encoding/test/readConditions.test.ts
sz-piotr: I am happy to change it but it is already like this everywhere. If you insist we can make a big update in another PR
image

sin6pi7: Ok, makes sense - let's do it in another PR at some point.

PR #89, opened 2022-04-01
Mapping between ETH addresses and Stark keys

sin6pi7:


packages/backend/src/peripherals/database/migrations/010_user_registration_events.ts
```diff
@@ -0,0 +1,27 @@
...
+If you find that something was incorrectly set up in the `up` function you
+should create a new migration file that fixes the issue.
+
+*/
+
+import { Knex } from 'knex'
+
+export async function up(knex: Knex) {
+  await knex.schema.createTable('user_registration_events', (table) => {
+    table.increments('id').primary(),
```
sz-piotr, line 18: what is the `,` for?

sz-piotr (approved):

PR #92, opened 2022-04-05
Include state updates without balance updates in the main list

sin6pi7:


packages/backend/test/peripherals/database/StateUpdateRepository.test.ts
```diff
@@ -206,25 +206,36 @@ describe(StateUpdateRepository.name, () => {
         prices: [],
       })
     }
+    await repository.add({
```
sz-piotr, line 209: This entry is not present in the expected output in the test

sz-piotr (changes requested):

sz-piotr (approved):

PR #93, opened 2022-04-05
Unify timestamp

sin6pi7:
- Add Timestamp type
- Use Timestamp in backend
- Use Timestamp in state and encoding
- Final touches


packages/backend/src/peripherals/database/RollupStateRepository.ts
```diff
@@ -131,7 +131,7 @@ export class RollupStateRepository implements IRollupStateStorage {
 
 function parametersToJson(parameters: RollupParameters) {
   return {
-    timestamp: parameters.timestamp.toString(),
+    timestamp: BigInt(Number(parameters.timestamp)),
```
sz-piotr, line 134: There are no bigints in json. This has to be a number then

packages/encoding/test/decodeOnChainData.test.ts
```diff
@@ -5,7 +5,7 @@ import REAL_DECODED from './data/onchain-decoded.json'
 import REAL_DATA from './data/onchain-example.json'
 
 describe('decodeOnChainData', () => {
-  it('decodes real onchain data', () => {
+  it('decodes real onchain data', async () => {
```
sz-piotr, line 8: why?

packages/state/src/RollupState.ts
```diff
@@ -57,7 +57,7 @@ export class RollupState {
           throw new Error('Invalid update count')
         }
         const funding =
-          update.fundingTimestamp !== 0n
+          Number(update.fundingTimestamp) !== 0
```
sz-piotr, line 60: Maybe `Timestamp(0)` would be cleaner here?

sz-piotr (changes requested):

packages/backend/src/peripherals/database/RollupStateRepository.ts
sin6pi7: Returned type is not really a json - `timestamp` is of type `bigint` and `funding` of `jsonb`. I think naming here is confusing, so I changed `json` to `row`, which reflects reality.

packages/encoding/test/decodeOnChainData.test.ts
sin6pi7: That's on me - forgot to clean this up - removed now.

packages/state/src/RollupState.ts
sin6pi7: You're right - done.

sz-piotr (approved):

PR #95, opened 2022-04-13
Forced actions

sin6pi7:


packages/backend/src/core/ForcedEventsCollector.ts
```diff
@@ -0,0 +1,103 @@
...
+              eventType: 'mined' as const,
+              blockNumber,
+              transactionHash,
+              timestamp,
+              publicKeyA: event.args.starkKeyA.toHexString(),
+              publicKeyB: event.args.starkKeyB.toHexString(),
+              positionIdA: BigInt(event.args.vaultIdA),
+              positionIdB: BigInt(event.args.vaultIdB),
+              syntheticAssetId: decodeAssetId(
+                event.args.syntheticAssetId.toHexString().slice(2)
```
sin6pi7, line 76: @sz-piotr take a look at this please

packages/backend/src/core/StateUpdateCollector.ts
```diff
@@ -77,28 +85,75 @@ export class StateUpdateCollector {
...
+
+  async processForcedActions(
+    forcedActions: ForcedAction[],
+    timestamp: Timestamp,
+    blockNumber: number,
+    stateUpdateId: number
+  ): Promise<void> {
+    const datas = forcedActions.map((action) =>
+      action.type === 'trade'
+        ? omit(action, 'type', 'nonce')
```
sin6pi7, line 124: @sin6pi7 this might be needed for hash calculations - check and update if necessary

packages/backend/src/peripherals/database/ForcedTransactionsRepository.ts
```diff
@@ -0,0 +1,419 @@
...
+    }
+  }, {})
+}
+
+function toJsonString(data: object) {
+  return JSON.stringify(toSerializableJson(data))
+}
+
+function hashData(data: string): string {
+  return createHash('md5').update(data).digest('base64')
```
sin6pi7, line 91: @sin6pi7 What's the collision probability here? Maybe something better

packages/backend/src/peripherals/database/ForcedTransactionsRepository.ts
```diff
@@ -0,0 +1,419 @@
...
+
+  async getTransactionHashesByMinedEventsData(
+    datas: (TradeMinedEventData | WithdrawalMinedEventData)[]
+  ): Promise<(Hash256 | undefined)[]> {
+    if (datas.length === 0) {
+      return []
+    }
+    const hashes = datas.map(toJsonString).map(hashData)
+    const events = await this.knex('forced_transaction_events').whereRaw(
+      "data->>'hash' = any(?)",
```
sin6pi7, line 328: @sin6pi7 @sz-piotr what if there is more than one transaction with the same data hash?

packages/backend/src/peripherals/database/ForcedTransactionsRepository.ts
```diff
@@ -0,0 +1,419 @@
...
+  return Object.entries(data).reduce((acc, [key, val]) => {
+    return {
+      ...acc,
+      [key]: typeof val === 'bigint' ? val.toString() : val,
+    }
+  }, {})
+}
+
+function toJsonString(data: object) {
+  return JSON.stringify(toSerializableJson(data))
```
sin6pi7, line 87: @sin6pi7 key order matters

packages/backend/test/core/StateUpdateCollector.test.ts
```diff
@@ -196,4 +208,72 @@ describe(StateUpdateCollector.name, () => {
       ])
     })
   })
+
+  describe(StateUpdateCollector.prototype.processForcedActions.name, () => {
+    it('adds verified events only for existing mined events', async () => {
```
sin6pi7, line 213: @sin6pi7 shout about missing event

packages/frontend/src/pages/common/Table.tsx
```diff
@@ -11,22 +11,89 @@ type Row = {
 type Column = {
   header: string
   numeric?: boolean
-  maxWidth?: boolean
+  maxWidth?: 250 | 320
```
sin6pi7, line 14: @sin6pi7 consider passing classes instead of magic numbers

packages/frontend/src/pages/formatLargeNumber.ts
```diff
@@ -2,13 +2,16 @@ const units = ['', 'K', 'M', 'B', 'T']
 
 const HAIR_SPACE = '\u200a'
 
-export function formatLargeNumber(value: number): string {
+export function formatLargeNumber(value: number | bigint): string {
```
sin6pi7, line 5: @sin6pi7 @sz-piotr what is the input like?

packages/backend/src/peripherals/database/ForcedTransactionsRepository.ts
sin6pi7: @sz-piotr It is actually already taken care of - only one is returned (look at the `events.find` call in `hashes.map` function). However, the order of events is not defined - given two mined events with the same data shall we care about which one we choose? If so, what would be the logic?

packages/backend/src/core/StateUpdateCollector.ts
```diff
@@ -77,28 +84,71 @@ export class StateUpdateCollector {
...
+              timestamp,
+              blockNumber,
+              stateUpdateId,
+            }
+          : undefined
+      })
+      .filter(
+        (
+          e
+        ): e is
```
sz-piotr, line 142: `e is Exclude<typeof e, undefined>`

packages/backend/src/peripherals/database/ForcedTransactionsRepository.ts
```diff
@@ -0,0 +1,409 @@
...
+function toSerializableJson(data: object): json {
+  return Object.entries(data).reduce((acc, [key, val]) => {
+    return {
+      ...acc,
+      [key]: typeof val === 'bigint' ? val.toString() : val,
+    }
+  }, {})
+}
+
+function hashData(data: object): string {
```
sz-piotr, line 87: This function is a bit useless, because all it does it just calls another function

packages/backend/src/peripherals/database/types.ts
```diff
@@ -106,6 +107,17 @@ declare module 'knex/types/tables' {
     eth_address: string
   }
 
+  type ForcedTransactionEventRow = {
```
sz-piotr, line 110: As a general note in this PR. The rest of the codebase follows the style of declaring shapes as interfaces, but here and elsewhere in this PR type aliases are used. It would be cool to keep a consistent style

packages/frontend/src/pages/formatLargeNumber.ts
sz-piotr: Originally in L2BEAT the input is the USD TVL of something, e.g. 1234567890.1234 would be formatted as 1.23B

Because you do not multiply the bigint that you're passing you are actually formatting it incorrectly, so `1n` is going to be formatted as `0.01` instead of `1.00`

packages/backend/src/peripherals/database/types.ts
sin6pi7: Good point - changed.

Issue #97, opened 2022-04-26
Add search (KISS - just check format and build link)


sz-piotr:
Search options:
- state root hash - redirects to state update by id
- state transition fact hash - redirects to state update by id
- state update id - redirects to state update by id
- position id - redirects to position by id (problem: positions and state updates share same id space, maybe we should prefix, e.g. states are s1, s5123 and positions are p1, p5123123)
- ethereum address - redirects to position by id
- stark key - redirects to position by id
- transaction hash - redirects to forced transaction by hash
- offer id - redirects to offer id (same problem, maybe o1, o2, o123)
When nothing matches we display a 404 saying that the following search term was not found.

krzkaczor:
How to find these things?

1. state root hash:
Example: `0x00fdf1255546cbd21f78402e14b9db04c5740b4148f15d0e8782bf0d4732c38d`
How to get it? `state_updates.root_hash` -> get id -> redirect to state-updates/id
2. ethereum address:
Example: `0x7a51D0B0Ee843c572F197c63BbB8277A957c04c5`
How to get it? `user_registration_events.eth_address` -> get stark_key -> match via `positions.public_key`
3. stark key:
Example: `0x027cda895fbaa174bf10c8e0f57561fa9aa6a93cfec32b87f1bdfe55a161e358`
How to get it? similar to eth_address
4. transaction hash - transaction hash not available yet

PR #100, opened 2022-04-27
Basic search 

krzkaczor:
Basic search supporting querying via ethereum address or stark key or state_root_hash.

Closes: https://github.com/l2beat/dydx-state-explorer/issues/97


packages/backend/src/api/controllers/FrontendController.ts
```diff
@@ -355,4 +363,99 @@ export class FrontendController {
...
+    return {
+      status: 302,
+      url: `/positions/${positionId}`,
+    }
+  }
+}
+
+type Type = 'state_root_hash_or_stark_key' | 'ethereum_address' | 'unknown'
+
+function categorizeSearchQuery(query: string): Type {
```
sz-piotr, line 452: It could be easier if this method was names: `parseSearchQuery` and it returned a struct:
```ts
interface ParsedQuery {
  ethereumAddress: EthereumAddress | undefined,
  pedersenHash: PedersenHash | undefined,
  hash256: Hash256 | undefined,
  starkKey: string | undefined // TODO: type for StarkKey
  id: number | undefined
}
```
This way you preserve the static typing and enforce all potential constraints like ethereum address casing.
Then the code could just check for each option if it is undefined or not and try to return values accordingly.

sz-piotr (changes requested):

packages/backend/src/peripherals/database/StateUpdateRepository.ts
```diff
@@ -93,6 +93,24 @@ export class StateUpdateRepository {
     })
   }
 
+  async getPositionIdByRootHash(hash: string): Promise<bigint | undefined> {
```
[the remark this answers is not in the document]
krzkaczor: I think it's not PedersenHash but rather Hash256... idk I am really not sure about these terms...

sz-piotr (changes requested):
It's much better now but definitely missing an E2E test

packages/backend/src/peripherals/database/StateUpdateRepository.ts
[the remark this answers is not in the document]
sz-piotr: The root hash is a PedersenHash. The difference is that a PedersenHash is 31.5 bytes and so starts with a 0.

packages/backend/test/api/routers/FrontendRouter.test.ts
```diff
@@ -121,4 +135,97 @@ describe('FrontendRouter', () => {
       await server.get('/forced-transactions?page=foo&perPage=bar').expect(400)
     })
   })
+
+  describe('/search', async () => {
+    const { knex } = setupDatabaseTestSuite()
+    const ethAddress = EthereumAddress(
```
sz-piotr, line 141: EthereumAddress.random()

packages/backend/test/api/controllers/FrontendController.test.ts
```diff
@@ -0,0 +1,26 @@
+import { EthereumAddress, PedersenHash } from '@explorer/types'
+import { expect } from 'earljs'
+
+import { parseSearchQuery } from '../../../src/api/controllers/FrontendController'
+
+describe(parseSearchQuery.name, () => {
+  it('parses ethereum address', () => {
+    const ethAddress = '0xc730B028dA66EBB14f20e67c68DD809FBC49890D'
```
sz-piotr, line 8: EthereumAddress.random().toString()

sz-piotr (approved):

packages/backend/src/peripherals/database/StateUpdateRepository.ts
```diff
@@ -93,6 +93,26 @@ export class StateUpdateRepository {
     })
   }
 
+  async getPositionIdByRootHash(
```
sz-piotr, line 96: Missing test

packages/backend/src/peripherals/database/StateUpdateRepository.ts
```diff
@@ -93,6 +93,26 @@ export class StateUpdateRepository {
...
+    hash: PedersenHash
+  ): Promise<bigint | undefined> {
+    const rows = await this.knex('state_updates')
+      .where('root_hash', hash.toString())
+      .select('state_updates.id as id')
+
+    return rows[0]?.id
+  }
+
+  async getPositionIdByPublicKey(
```
sz-piotr, line 106: Missing test

packages/backend/src/peripherals/database/UserRegistrationEventRepository.ts
```diff
@@ -80,4 +80,15 @@ export class UserRegistrationEventRepository {
       .limit(1)
     return rows.length > 0 ? toRecord(rows[0]) : undefined
   }
+
+  async findByEthereumAddress(
```
sz-piotr, line 84: Missing test

sz-piotr (changes requested):

packages/backend/test/peripherals/database/UserRegistrationEventRepository.test.ts
```diff
@@ -139,4 +139,34 @@ describe(UserRegistrationEventRepository.name, () => {
...
+        ethAddress: EthereumAddress.ZERO,
+        starkKey: '0x1234',
+      },
+      event,
+    ]
+
+    await repository.add(records)
+    const actual = await repository.findByEthereumAddress(event.ethAddress)
+
+    expect(actual).toEqual({ id: 2, ...event })
```
sz-piotr, line 162: This test is failing with id being `11` and not `2`. Are the ids here even something we're concerned about? I think in other places we use `expect.a(Number)`.

I also checked that this test was failing before I merged the master branch into this branch.

sz-piotr (changes requested):
I have refactored the FrontendController and so when I merged master into your branch I created a new class called SearchController. The code is basically the same, just situated in a different place.

packages/backend/src/api/controllers/SearchController.ts
```diff
@@ -0,0 +1,113 @@
...
+    return {
+      stateTreeHash: PedersenHash(query),
+      starkKey: StarkKey(query),
+    }
+  }
+  if (query.startsWith('0x') && query.length === 42) {
+    return {
+      ethereumAddress: EthereumAddress(query),
+    }
+  }
```
sz-piotr, line 110: Because `PedersenHash`, `StarkKey` and `EthereumAddress` can throw and you're partially duplicating their logic here I think it's better to use the following structure:
```ts
const parsed: ParsedQuery = {}
try {
  parsed.stateTreeHash: PedersenHash(query),
} catch {}
try {
  parsed.starkKey: StarkKey(query),
} catch {}
try {
  parsed.ethereumAddress: EthereumAddress(query),
} catch {}
return parsed
```

sz-piotr (changes requested):

packages/backend/src/api/controllers/SearchController.ts
krzkaczor: Good idea! I refactored it using `tryOrUndefined`. WDYT? is there a good place for such util?

sz-piotr: Below is great

sz-piotr (approved):

PR #103, opened 2022-04-29
Improve navigation between projects

krzkaczor:
This should allow clicking on symbols imported from a different, internal package and navigating to the sourcecode rather than to `.d.ts` file. 

Goes best with additional vscode settings:
```
// https://twitter.com/krzKaczor/status/1120631267781951488
  "editor.gotoLocation.multipleDefinitions": "goto",
  "editor.gotoLocation.multipleImplementations": "goto",
  "editor.gotoLocation.multipleTypeDefinitions": "goto",
  "editor.gotoLocation.multipleDeclarations": "goto",
  "editor.gotoLocation.multipleReferences": "goto"
```

### DIFF
--- a/packages/backend/tsconfig.json
+++ b/packages/backend/tsconfig.json
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "outDir": "build",
-    "declaration": false
+    "declaration": false,
+    "declarationMap": false
   }
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -10,6 +10,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "react",
-    "declaration": true
+    "declaration": true,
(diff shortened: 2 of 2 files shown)


packages/backend/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "outDir": "build",
-    "declaration": false
+    "declaration": false,
+    "declarationMap": false
```
sin6pi7, line 6: Why is it necessary to make this `false` for `backend` package?

sin6pi7: Aight, that's something we can improve later on if necessary.

sz-piotr: That's weird, I can't remember why this was done. I think it's fine to enable them

sz-piotr (approved):

PR #115, opened 2022-05-06
Add force transaction form UI

sz-piotr:
This PR only adds the visual UI part without any logic or interactions


krzkaczor:
I am having a hard time trying to run it locally. The form should be available at: `/forced-transactions/form`, no?
But for some reason I am getting `{"issues":[{"code":"custom","message":"Invalid input","path":["params","hash"]}]}`.

I tried rebuilding frontend project with `yarn build`. Btw. is there a watchmod for frontend? `yarn dev` seems to break the build...

sz-piotr:
@krzkaczor 

To run it locally you need to use the `yarn dev` in frontend, because it is not implemented on the backend

michalsidzej (commented):
LGTM

PR #116, opened 2022-05-06
Add offers to DB

michalsidzej:
This PR doesn't include functionality that allows mark an offer as submitted, it will be added when [this](https://github.com/l2beat/dydx-state-explorer/pull/122) is merged


packages/backend/README.md
```diff
@@ -37,13 +29,14 @@ docker exec -it state_explorer_postgres psql -U postgres -c 'CREATE DATABASE loc
...
-If you restart your system running `docker start state_explorer_postgres` will bring the database back online.
+If you restart your system, running `docker start state_explorer_postgres` will bring the database back online.
 
-Once you have everything create a `.env` file with the following contents:
+Once you have everything, create a `.env` file with the following contents:
 
 ```
 LOCAL_DB_URL=postgresql://postgres:password@localhost:5432/local
 TEST_DB_URL=postgresql://postgres:password@localhost:5432/test
+LOCAL_JSON_RPC_URL=https://eth-mainnet.alchemyapi.io/v2/zR5UQGm__S5p1pVRUG3vCsK5elOgJL6v
```
sz-piotr, line 39: Don't push secrets to the repository :) Instead you can say `<PASTE_YOUR_ALCHEMY_URL_HERE>`

packages/backend/src/api/controllers/ControllerResult.ts
```diff
@@ -17,3 +17,7 @@ export interface ControllerRedirectResult {
   type: 'redirect'
   url: string
 }
+
+export interface ControllerCreatedResult {
```
sz-piotr, line 21: You can add it to the discriminated union `ControllerResult` on the top of the file

packages/backend/src/api/controllers/OfferController.ts
```diff
@@ -0,0 +1,13 @@
+import {
+  OfferRecord,
+  OfferRepository,
+} from '../../peripherals/database/OfferRepository'
+
+export class OfferController {
+  constructor(private offerRepository: OfferRepository) {}
+
+  async postOffer(offer: OfferRecord) {
+    await this.offerRepository.add([offer])
```
sz-piotr, line 10: We want some validation here that the offer actually makes sense given the user's position

packages/backend/src/api/routers/OffersRouter.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export function createOffersRouter(offerController: OfferController) {
+  const router = new Router()
+
+  router.post(
+    '/offer',
+    bodyParser(),
+    withTypedContext(
+      z.object({
+        body: z.object({
```
sz-piotr, line 17: This won't make Typescript happy. It's best to nest it under request here.

packages/backend/src/peripherals/database/OfferRepository.ts
```diff
@@ -0,0 +1,74 @@
...
+  amount_synthetic: bigint
+  a_is_buying_synthetic: boolean
+}
+
+export class OfferRepository implements Repository<OfferRecord> {
+  constructor(private knex: Knex, private logger: Logger) {
+    this.logger = logger.for(this)
+  }
+
+  async add(offers: OfferRecord[]) {
```
sz-piotr, line 30: We'll probably only be adding one offer at a time. Also, since we're using synthetic ids it's worthwhile to return the id from the method and also alongside 201 created

packages/backend/src/peripherals/database/OfferRepository.ts
```diff
@@ -0,0 +1,74 @@
...
+      return
+    }
+
+    const rows: OfferRow[] = offers.map(toRow)
+    await this.knex('offers').insert(rows)
+
+    this.logger.debug({ method: 'add', rows: rows.length })
+  }
+
+  async getAll(): Promise<OfferRecord[]> {
```
sz-piotr, line 42: We also want a findById method

sz-piotr (changes requested):

packages/backend/src/peripherals/database/migrations/013_forced_trade_offers.ts
```diff
@@ -0,0 +1,51 @@
...
+because migrations are only run once on the production server. 
+
+If you find that something was incorrectly set up in the `up` function you
+should create a new migration file that fixes the issue.
+
+*/
+
+import { Knex } from 'knex'
+
+export async function up(knex: Knex) {
```
sin6pi7, line 16: Why two separate tables instead of one?

packages/backend/src/api/controllers/ForcedTradeOfferController.ts
```diff
@@ -0,0 +1,289 @@
...
+
+export class ForcedTradeOfferController {
+  constructor(
+    private offerRepository: ForcedTradeOfferRepository,
+    private stateUpdateRepository: StateUpdateRepository,
+    private userRegistrationEventRepository: UserRegistrationEventRepository
+  ) {}
+
+  async postOffer(
+    offer: Omit<ForcedTradeInitialOfferRecord, 'createdAt' | 'id'>
```
sz-piotr, line 29: Not in scope of this PR, but I would add a requirement for the user to also sign this offer so that only owners of the positions can create offers.

packages/backend/src/peripherals/database/migrations/013_forced_trade_offers.ts
michalsidzej: It's easier to make it type-safe from my perspective. If it were one DB, half of the fields would be nullable. With this approach, all the necessary fields are staying required.

packages/backend/src/api/controllers/ForcedTradeOfferController.ts
michalsidzej: Ok, but should we merge this first, as the scope is getting quite broad, it will be harder to do code review, what do you think?

packages/backend/src/api/controllers/ForcedTradeOfferController.ts
```diff
@@ -0,0 +1,289 @@
...
+    const acceptedAt = Timestamp(Date.now())
+    await this.offerRepository.addAcceptOffer(initialOfferId, {
+      acceptedAt,
+      ...acceptOfferData,
+    })
+
+    return { type: 'success', content: 'Accept offer was submitted.' }
+  }
+
+  async calculateFrozenBalance(starkKey: StarkKey, assetId: AssetId) {
```
sz-piotr, line 121: I like the idea of this but this calculation actually has two problems:
1. I think that it makes sense for a user to create multiple offers for the same asset at different prices / amounts. Even if they only have 1000 units they should be able to create offers up to that amount.
2. When selling / buying user's USDC balance doesn't matter. Trading usually does not affect the total value of an account.

packages/backend/src/api/controllers/ForcedTradeOfferController.ts
```diff
@@ -0,0 +1,289 @@
...
+    )
+
+    if (accceptOffer) {
+      return {
+        type: 'bad request',
+        content: 'Offer already accepted by a user.',
+      }
+    }
+
+    const frozenBalance = await this.calculateFrozenBalance(
```
sz-piotr, line 93: When someone agrees to trade they don't actually need to have any of the assets that the other party wants.

packages/backend/src/api/routers/ForcedTradeOfferRouter.ts
```diff
@@ -0,0 +1,72 @@
...
+    withTypedContext(
+      z.object({
+        params: z.object({
+          initialOfferId: stringAsInt(),
+        }),
+        request: z.object({
+          body: z.object({
+            starkKeyB: stringAs(StarkKey),
+            positionIdB: stringAsBigInt(),
+            submissionExpirationTime: stringAsInt(), // Timestamp?
```
sz-piotr, line 55: Actually I think this is in hours? Please double check with solidity code

packages/backend/src/api/routers/ForcedTradeOfferRouter.ts
```diff
@@ -0,0 +1,72 @@
...
+      }),
+      async (ctx) => {
+        const result = await offerController.postOffer(ctx.request.body)
+        applyControllerResult(ctx, result)
+      }
+    )
+  )
+
+  router.post(
+    '/offer/:initialOfferId/accept',
```
sz-piotr, line 44: `router.put('offer/:initialOfferId')`, we use PUT, because we modify the resource at this url

packages/backend/test/api/routers/utils.test.ts
```diff
@@ -0,0 +1,67 @@
...
+const REDIRECT_RESULT: ControllerRedirectResult = {
+  type: 'redirect',
+  url: 'url',
+}
+
+describe(applyControllerResult.name, () => {
+  const ctx = mock<Context>({
+    redirect: (url) => url,
+  })
+  it('handles success result', () => {
```
sz-piotr, line 43: Nitpick. We like newlines between `it` calls

sz-piotr (changes requested):

packages/backend/src/api/controllers/ForcedTradeOfferController.ts
michalsidzej: So actually it's not necessary to check the `frozenBalance`

packages/backend/src/api/routers/ForcedTradeOfferRouter.ts
michalsidzej: Yes, seems like https://github.com/starkware-libs/starkex-contracts/blob/master/scalable-dex/contracts/src/perpetual/interactions/ForcedTrades.sol#L57 Interestingly, the [tx](https://etherscan.io/tx/0x9b2dce5538d0c8c08511c9383be9b67da6f952b367baff0c8bdb5f66c9395634) that we use as an example, had an expiration time in 40k years

packages/backend/src/peripherals/database/migrations/013_forced_trade_offers.ts
sin6pi7: As long as I agree with you in general, in this case it results in more database and application code where at the same time we validate the correctness of data at the application level already.

sz-piotr (approved):

PR #118, opened 2022-05-09
Handle user input in the form

sz-piotr:
This PR still doesn't implement actually sending transactions


packages/frontend/src/scripts/transaction/getFormElements.ts
```diff
@@ -0,0 +1,32 @@
...
+    priceInput: $<HTMLInputElement>('#price'),
+    suggestedPriceView: $('#asset-price'),
+    suggestedPriceButton: $<HTMLButtonElement>('#suggested-price'),
+    totalSection: $('#total-section'),
+    totalInput: $<HTMLInputElement>('#total'),
+    submitButton: $<HTMLButtonElement>('#submit'),
+    exitButton: $<HTMLButtonElement>('#exit-button'),
+    buyButton: $<HTMLButtonElement>('#buy-button'),
+    sellButton: $<HTMLButtonElement>('#sell-button'),
+    infoSection: $('#info-section'),
```
sin6pi7, line 30: Wdyt about moving these ids to an external module and import it both here and in views? That way there is a smaller chance of making a typo

sin6pi7 (commented):
😍 apart from shared id strings LGTM!

sin6pi7 (approved):

PR #119, opened 2022-05-10
Tidy up contract addresses

sin6pi7:


packages/backend/src/peripherals/ethereum/addresses.ts
```diff
@@ -0,0 +1,3 @@
+export const PERPETUAL_ADDRESS = '0xD54f502e184B6B739d7D27a6410a67dc462D69c8'
+export const REGISTRY_ADDRESS = '0xEfbCcE4659db72eC6897F46783303708cf9ACef8'
+export const PROXY_ADDRESS = '0xC8c212f11f6ACca77A7afeB7282dEBa5530eb46C'
```
sz-piotr, line 3: This is a good first step. Ideally all the addresses would be injected into the classes that use them, so that the system can run on different networks with different addresses

sz-piotr (approved):

PR #122, opened 2022-05-10
Watch transaction status

sin6pi7:
TODO:

- [ ] tests


packages/backend/src/core/ForcedTransactionMonitor.ts
```diff
@@ -0,0 +1,97 @@
...
+      console.error(error) // TODO: log properly
+      return null
+    }
+  }
+
+  private async waitForReceipt(hash: Hash256) {
+    return promiseRetry(
+      (retry) =>
+        this.ethereumClient.waitForTransactionReceipt(hash).catch(retry),
+      { forever: true }
```
sz-piotr, line 47: This also shouldn't be forever

packages/backend/src/core/ForcedTransactionMonitor.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  async watchTransaction(tx: ForcedTransaction) {
+    const transaction = await this.waitForTransaction(tx.hash)
+    if (!transaction) {
+      await this.forcedTransactionRepository.delete(tx.hash)
+      return
+    }
+    const receipt = await this.waitForReceipt(tx.hash)
+    if (receipt.status === 0) {
+      await this.forcedTransactionRepository.delete(tx.hash)
```
sz-piotr, line 59: Why delete when reverted? We should show the reverted status.

packages/backend/src/core/ForcedTransactionMonitor.ts
```diff
@@ -0,0 +1,97 @@
...
+            eventType: 'mined' as const,
+            blockNumber,
+            timestamp,
+            transactionHash,
+            publicKey: tx.publicKey,
+            positionId: tx.positionId,
+            amount: tx.amount,
+          }
+
+    await this.forcedTransactionRepository.addEvents([event])
```
sz-piotr, line 95: Personally I am not sure I like this design. Watching transactions seem like an activity that could be done in an abstract way. So you can have ForcedTransactionRepository but you can also have a TransactionStatusRepository which has:

- hash
- time created
- forgotten - boolean
- mined - boolean
- reverted - boolean
- time mined

Then when constructing the history you would just join the two tables. This decouples watching transactions from forced transactions and makes testing easier.

packages/backend/src/core/ForcedTransactionMonitor.ts
sin6pi7: Ah, you're right.

packages/backend/src/core/ForcedTransactionMonitor.ts
sin6pi7: `waitForTransactionReceipt` from `ethers` waits for a transaction until it is mined and it should not throw unless something unexpected happens such as network error. My reasoning was that we'd rather wait longer than consider a mined/reverted transaction forgotten, because of something unexpected. Taking that into consideration I went with `{ forever: true }`, because I was not sure how long we should wait to safely consider a transaction forgotten in such case.

PR #124, opened 2022-05-12
Unify repository code

sz-piotr:
This PR does the following things:
1. Unify naming across repositories
2. Enforce consistent logging
3. Remove the type argument from KeyValueStore
4. Extract PositionRepository from StateUpdateRepository


michalsidzej (approved):
LGTM

PR #125, opened 2022-05-12
Reorganise forced transactions

sin6pi7:


packages/backend/src/peripherals/database/migrations/014_forced_transactions.ts
```diff
@@ -0,0 +1,36 @@
...
+because migrations are only run once on the production server. 
+
+If you find that something was incorrectly set up in the `up` function you
+should create a new migration file that fixes the issue.
+
+*/
+
+import { Knex } from 'knex'
+
+import { up as createForcedTransactionEvents } from './012_forced_transaction_events'
```
sz-piotr, line 16: Is there a reason up is used, but down is reimplemented with `dropTableIfExists`?

sin6pi7: Good catch - removed.

packages/backend/src/peripherals/database/ForcedTransactionsRepository.ts
```diff
@@ -1,351 +1,150 @@
...
+export class ForcedTransactionsRepository extends BaseRepository {
+  constructor(knex: Knex, logger: Logger) {
+    super(knex, logger)
+    this.getAll = this.wrapGet(this.getAll)
+    this.getLatest = this.wrapGet(this.getLatest)
+    this.getIncludedInStateUpdate = this.wrapGet(this.getIncludedInStateUpdate)
+    this.getAffectingPosition = this.wrapGet(this.getAffectingPosition)
+    this.getAffectingPosition = this.wrapGet(this.getAffectingPosition)
+    this.findByHash = this.wrapFind(this.findByHash)
+    this.deleteAll = this.wrapDelete(this.deleteAll)
```
sz-piotr, line 121: What about `add` and `markAsMined`?

packages/backend/src/peripherals/database/migrations/014_forced_transactions.ts
sz-piotr: Why not just import down too?

sin6pi7: Another good catch :D

sz-piotr (approved):

PR #126, opened 2022-05-13
Serve /forced/new on the backend

sz-piotr:


packages/backend/src/peripherals/database/BaseRepository.ts
```diff
@@ -37,25 +37,19 @@ export class BaseRepository {
...
   ): AnyMethod<A, R> {
-    return async (...args: A) => {
-      const result = await method.call(this, ...args)
-      this.logger.debug({ method: method.name })
-      return result
-    }
+    return this.wrap(method, () => this.logger.debug({ method: method.name }))
   }
 
   protected wrapAdd<T, R extends number | string>(
```
sin6pi7, line 43: Will this work for add methods returning domain ids such as `Hash256`?

sin6pi7 (approved):
Apart from BaseRepository signatures (which is a separate concern than this PR) it all LGTM.

PR #130, opened 2022-05-17
Fix invalid event decoding

sz-piotr:


packages/backend/src/Application.ts
```diff
@@ -201,8 +201,10 @@ export class Application {
       await databaseService.migrateToLatest()
 
       await apiServer.listen()
-      await syncScheduler.start()
-      await blockDownloader.start()
+      if (config.enableSync) {
```
sin6pi7, line 204: What is the reason for introducing this?

sz-piotr: I wanted to connect to production database locally but I also didn't want to modify it

sz-piotr: I can extract this to a separate PR if you want

sin6pi7: No, that's fine to be here - simple enough to understand

sin6pi7 (approved):

PR #131, opened 2022-05-17
Watch transaction status

sin6pi7:


packages/backend/src/core/TransactionStatusMonitor.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  start() {
+    this.scheduleNextCheck()
+  }
+
+  stop() {
+    if (!this.timeout) {
+      return
+    }
+    clearTimeout(this.timeout)
```
sz-piotr, line 28: This doesn't work if we're in the middle of execution. An easier way is to have a member `running: boolean` and `scheduleNextCheck` just checks this and the callback in setTimeout does too

packages/backend/src/Application.ts
```diff
@@ -201,6 +216,7 @@ export class Application {
       await databaseService.migrateToLatest()
 
       await apiServer.listen()
+      transactionStatusMonitor.start()
```
sz-piotr, line 219: Move below enableSync

packages/backend/src/peripherals/database/BaseRepository.ts
```diff
@@ -40,11 +40,12 @@ export class BaseRepository {
...
   }
 
-  protected wrapAdd<T, R extends number | string>(
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  protected wrapAdd<T, R extends number | string | String | Number>(
     method: AddMethod<T, R>
   ): AddMethod<T, R> {
     return this.wrap(method, (id) =>
-      this.logger.debug({ method: method.name, id })
+      this.logger.debug({ method: method.name, id: id.toString() })
```
sz-piotr, line 48:
```suggestion
      this.logger.debug({ method: method.name, id: id.valueOf() })
```
This actually works for numbers

packages/backend/src/peripherals/database/TransactionStatusRepository.ts
```diff
@@ -0,0 +1,122 @@
...
+function timestampToBigInt(timestamp?: Nullable<Timestamp>): Nullable<bigint> {
+  return !timestamp ? null : BigInt(timestamp.toString())
+}
+
+function toPartialRow(
+  record: Partial<Record> & { hash: Record['hash'] }
+): Partial<Row> & { hash: Row['hash'] } {
+  const result: Partial<Row> & { hash: Row['hash'] } = {
+    hash: record.hash.toString(),
+    sent_at: record.sentAt && timestampToBigInt(record.sentAt),
```
sz-piotr, line 58: This and the following checks don't work for `0`. In this case maybe it's acceptable, because we won't have timestamps from 1970, but in general it is an error. I'd be happier with explicit undefined checks

packages/backend/src/peripherals/database/TransactionStatusRepository.ts
```diff
@@ -0,0 +1,122 @@
...
+    const deleted = await this.knex('transaction_status').delete()
+    return deleted
+  }
+
+  async getWaitingToBeMined(): Promise<Record[]> {
+    const rows = await this.waitingToBeMinedQuery()
+    return rows.map(toRecord)
+  }
+
+  async updateWaitingToBeMined(
```
sz-piotr, line 113: Maybe `updateIfWaitingToBeMined` is better? I was confused because I thought that this function actually reset the tx to a state where it's waiting to be mined.

sz-piotr (changes requested):

sz-piotr (approved):

PR #133, opened 2022-05-18
Add initial offer signing and improve OfferController tests

michalsidzej:


packages/backend/src/api/controllers/ForcedTradeOfferController.ts
```diff
@@ -152,7 +167,24 @@ function validateBalance(
   return false
 }
 
-export function validateSignature(
+export function validateInitialSignature(
+  offer: Omit<ForcedTradeInitialOfferRecord, 'createdAt' | 'id'>,
+  signature: string,
+  ethAddressA: EthereumAddress
+) {
+  const stringOffer = JSON.stringify(
```
sz-piotr, line 175: I'd prefer being explicit here with:
```ts
JSON.stringify({
  field1: value1,
  field2: value2,
  ...
})
```
This way the field order is clear and the transformation can be done in place too.

packages/frontend/src/scripts/transaction/sign.ts
```diff
@@ -0,0 +1,25 @@
...
+  offer: ForcedTradeInitialOfferEntry,
+  address: EthereumAddress
+) {
+  const provider = window.ethereum
+
+  if (!provider || !address) {
+    return
+  }
+
+  const stringOffer = JSON.stringify(offer, null, 2)
```
sz-piotr, line 15: Again, I'd prefer explicit here, because then the field order (and count, as TS doesn't prevent additional fields) is guaranteed.

packages/frontend/src/scripts/transaction/submit.ts
```diff
@@ -0,0 +1,47 @@
+import { EthereumAddress } from '@explorer/types'
+
+import { signInitial } from './sign'
+import { FormState } from './types'
+
+export async function submit(state: FormState) {
```
sz-piotr, line 6: Outside the scope of this PR but proper states in the UI would be helpful here, e.g. loading, error, waiting for signature

packages/frontend/src/scripts/transaction/types.ts
```diff
@@ -73,3 +73,11 @@ export interface ModifyTotalAction {
   type: 'ModifyTotal'
   value: string
 }
+export interface ForcedTradeInitialOfferEntry {
```
sz-piotr, line 76: newline before

sz-piotr (approved):

PR #136, opened 2022-05-19
Homepage latest offer

sin6pi7:
Some remarks:
1. Buy/Sell buttons are not styled the way it is designed in figma, because of overlapping blue color on hover - this needs to be thought through, so as of now the best we can do is to keep these as simple as possible,
2. Database types for offers are lacking - adding them results in quite a few required updates in the repository class, so in order to limit the scope of this PR I decided to add them in another one.


packages/frontend/src/pages/home/Home.tsx
```diff
@@ -96,12 +98,48 @@ export function Home(props: HomeProps) {
...
+        ]}
+        rows={props.forcedTradeOffers.map((offer) => {
+          const link = `/forced/offers/${offer.id}`
+          return {
+            link,
+            cells: [
+              offer.type,
+              <AssetCell assetId={offer.assetId} amount={offer.amount} />,
+              formatCurrencyUnits(offer.price, 'USD'),
+              formatCurrencyUnits(offer.total, AssetId.USDC),
```
michalsidzej, line 137: I think that the designs imply displaying the price in USD in both cases. But we are storing the amounts in USDC, so it should probably look sth like this 
```suggestion
              formatCurrency(offer.price / 1000n, 'USD'),
              formatCurrency(offer.total / 1000n, 'USD'),
```
Price is also formatted like this, so there should not be a problem with precision.

michalsidzej: It doesn't look right rn, the numbers don't really add up
Screenshot 2022-05-19 at 11 17 32

michalsidzej (changes requested):

packages/backend/src/api/controllers/utils/toForcedTradeOfferEntry.ts
```diff
@@ -0,0 +1,32 @@
+import { ForcedTradeOfferEntry } from '@explorer/frontend'
+import { AssetId } from '@explorer/types'
+
+import {
+  ForcedTradeInitialOfferRecord,
+  ForcedTradeOfferRecord,
+} from '../../../peripherals/database/ForcedTradeOfferRepository'
+
+function getPriceInUSDCents(
```
michalsidzej, line 9: Would be good to see a test of this fn, decimals are hard.

packages/frontend/src/pages/common/table/Table.tsx
```diff
@@ -11,21 +11,23 @@ export function Table({ columns, rows, className, noRowsText }: TableProps) {
...
-              </th>
-            ))}
+            {columns.map(
+              ({ header, numeric, fullWidth, textAlignClass }, i) => (
+                <th
+                  scope="col"
+                  key={i}
+                  className={cx(
+                    'p-px pb-0.5 first:pl-0 last:pr-0',
+                    textAlignClass || (numeric ? 'text-right' : 'text-left'),
```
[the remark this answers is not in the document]
sin6pi7: This was a design decision - it is like that in all the tables throughout our designs. The only case where we have numeric content and the text is not aligned right is this one.

packages/frontend/src/pages/common/table/types.ts
```diff
@@ -10,6 +10,7 @@ export interface Column {
   numeric?: boolean
   monospace?: boolean
   fullWidth?: boolean
+  textAlignClass?: 'text-left' | 'text-right'
```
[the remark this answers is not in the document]
sin6pi7: Numeric brings a default align, which was a design decision. This allows for a combination of numeric styles and different text align, which is required for latest offers table in particular.

packages/frontend/src/pages/home/Home.tsx
sin6pi7: @sz-piotr mentioned that price should be displayed in USD and total in USDC - designs are not correct anymore if that decision stays.

packages/frontend/src/pages/common/table/Table.tsx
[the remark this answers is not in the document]
michalsidzej: Oh, that's the case then. But this is not purely numeric, as it also holdes information about the type of asset. I think it makes sense to just call it not numeric

michalsidzej (approved):

PR #138, opened 2022-05-19
Sign accept offer

michalsidzej:


sin6pi7 (approved):

packages/frontend/src/scripts/offer/sign.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  const actionHash = keccak256(
+    ['string', 'bytes'],
+    ['FORCED_TRADE', packedParameters]
+  )
+
+  const dataHashToSign = keccak256(
+    ['bytes32', 'uint256'],
+    [actionHash, accepted.submissionExpirationTime]
+  )
```
sin6pi7, line 83: The part of preparing parameters (i.e. packing and keccak'ing) looks pretty similar to what we do on the backend. Same for initial offer. As far as my understanding goes we do this to 1. create signatures on the frontend, 2. validate them on the backend. If that's the case extracting the part of preparing parameters would lower the chance of making an error and deduplicate the code.

sin6pi7:
Awesome 👍

PR #139, opened 2022-05-19
List all offers (HTML)

michalsidzej:


packages/backend/src/api/controllers/ForcedTradeOfferController.ts
```diff
@@ -26,6 +28,17 @@ export class ForcedTradeOfferController {
     private userRegistrationEventRepository: UserRegistrationEventRepository
   ) {}
 
+  async getOffersIndexPage(
+    account: EthereumAddress | undefined
+  ): Promise<ControllerResult> {
+    const offers = await this.offerRepository.getAllInitialOffers()
```
sin6pi7, line 34: We probably want to paginate this (and filter by type and asset), but this PR is about the HTML, so I'm just leaving it here as a note.

packages/frontend/src/preview/data.ts
```diff
@@ -62,9 +63,11 @@ const createFakeOffers = (count: number): ForcedTradeOfferEntry[] =>
       (amountCollateral * BigInt(10 ** AssetId.decimals(assetId))) /
       amountSynthetic /
       10_000n
+    const createdAt = Timestamp(Date.now() - 50000 - i * 480 * 3570)
```
sin6pi7, line 66: Just out of curiosity - is there any reasoning for these numbers?

michalsidzej: Just wanted to see something that's not regularly every 30 minutes

packages/backend/src/api/controllers/ForcedTradeOfferController.ts
```diff
@@ -29,12 +29,22 @@ export class ForcedTradeOfferController {
...
+    perPage: number,
     account: EthereumAddress | undefined
   ): Promise<ControllerResult> {
-    const offers = await this.offerRepository.getInitial()
+    const offers = await this.offerRepository.getLatestInitial({
+      offset: (page - 1) * perPage,
+      limit: perPage,
+    })
+
+    const total = await this.offerRepository.initialCount()
```
sin6pi7, line 41: You can make this run in parallel i.e.
```
    const [total, offers] = await Promise.all([
        this.offerRepository.initialCount(),
        this.offerRepository.getLatestInitial({
            offset: (page - 1) * perPage,
            limit: perPage,
        })
    ])
```

sin6pi7 (approved):

PR #140, opened 2022-05-19
Connect offers and transactions in database

sin6pi7:
Adding transaction hash to `forced_trade_offers` is enough - no need to add anything else to other tables.

### DIFF
--- a/packages/backend/src/peripherals/database/BaseRepository.ts
+++ b/packages/backend/src/peripherals/database/BaseRepository.ts
@@ -92,8 +92,8 @@ export class BaseRepository {
   }
 
   protected wrapSave<T>(method: SaveMethod<T>): SaveMethod<T> {
-    return this.wrap(method, (count) =>
-      this.logger.debug({ method: method.name, count })
+    return this.wrap(method, (updated) =>
+      this.logger.debug({ method: method.name, updated })
     )
   }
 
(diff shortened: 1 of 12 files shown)


packages/backend/src/peripherals/database/ForcedTradeOfferRepository.ts
```diff
@@ -1,108 +1,135 @@
...
-    const rows = await this.knex('forced_trade_offers')
-      .whereNotNull('accepted_at')
-      .select('*')
-
-    return rows.map(acceptedOfferToRecord)
+      .where('id', '=', row.id)
+    return !!updates
   }
 
   async getLatest({
```
michalsidzej, line 135: Don't we need this to be the latest but not the accepted offers?

packages/backend/src/peripherals/database/BaseRepository.ts
```diff
@@ -26,6 +26,10 @@ interface DeleteMethod<A extends unknown[]> {
   (...args: A): Promise<number>
 }
 
+interface SaveMethod<T> {
```
michalsidzej, line 29: Maybe we can call it `UpdateMethod`? What do you think?

sin6pi7: update would take a partial - save updates the whole record.

packages/backend/src/peripherals/database/ForcedTradeOfferRepository.ts
sin6pi7: This makes more sense, good catch

michalsidzej (approved):
I like the updated naming, its cleaner

PR #144, opened 2022-05-24
Send forced exit transaction

sz-piotr:


packages/frontend/src/scripts/transaction/submit.ts
```diff
@@ -29,13 +66,8 @@ export async function submit(state: FormState) {
...
     },
-    body: JSON.stringify(
-      {
-        offer,
-        signature,
-      },
-      (_, value: unknown) =>
-        typeof value === 'bigint' ? value.toString() : value
+    body: JSON.stringify({ offer, signature }, (_, value: unknown) =>
+      typeof value === 'bigint' ? value.toString() : value
```
sin6pi7, line 70: `stringify`ing with `BigInt`s mapped to `string`s is also done `TransactionForm.tsx` - maybe it's worth it to put it into a separate util?

PR #145, opened 2022-05-24
Filter offers

sin6pi7:


packages/frontend/src/pages/offers/pagination/Pagination.tsx
```diff
@@ -0,0 +1,51 @@
...
+  }
+  if (assetId) {
+    params.append('assetId', assetId.toString())
+  }
+  const query = params.toString()
+  const baseUrl = `/forced/offers${query && `?${query}`}`
+
+  return (
+    <div className="relative pt-9">
+      <ServerPagination
```
sz-piotr, line 38: I am not sure I like the idea of ServerPagination wrapping other details. Pagination seems separate in a way, or at least in name. I get that this simplifies the js logic. Won't it work with two forms?

packages/frontend/src/pages/offers/ForcedTradeOffersIndexProps.ts
```diff
@@ -14,9 +16,12 @@ export interface ForcedTradeOfferEntry {
 export interface ForcedTradeOffersIndexProps {
   readonly account: EthereumAddress | undefined
   readonly offers: ForcedTradeOfferEntry[]
+  readonly assetIds: AssetId[]
   readonly params: {
     readonly perPage: number
     readonly page: number
+    readonly assetId?: AssetId
```
sz-piotr, line 23: This won't work. It needs to be `AssetId | 'all'`

packages/frontend/src/pages/offers/ForcedTradeOffersIndexProps.ts
```diff
@@ -14,9 +16,12 @@ export interface ForcedTradeOfferEntry {
...
+    readonly type?: OfferType
```
sz-piotr, line 24: This won't work. It needs to be `OfferType | 'all'`

packages/frontend/src/pages/common/pagination/ServerPagination.tsx
```diff
@@ -24,7 +26,9 @@ export function ServerPagination({
       page: page.toString(),
       perPage: perPage.toString(),
     })
-    return `${baseUrl}?${params}`
+    return baseUrl.includes('?')
+      ? `${baseUrl}&${params}`
+      : `${baseUrl}?${params}`
```
sz-piotr, line 31: To include previous params you should use URLSearchParams to parse the baseUrl and then construct new params that include the old ones.

packages/backend/test/peripherals/database/ForcedTradeOfferRepository.test.ts
```diff
@@ -96,16 +96,94 @@ describe(ForcedTradeOfferRepository.name, () => {
...
-    await repository.add(fakeInitialOffer())
-    expect(await repository.initialCount()).toEqual(1)
-
-    const initial = fakeInitialOffer()
-    const id = await repository.add(initial)
-    expect(await repository.initialCount()).toEqual(2)
-
-    await repository.save({ ...initial, id, accepted: fakeAccepted() })
-    expect(await repository.initialCount()).toEqual(1)
+  it('returns initial offers', async () => {
```
sz-piotr, line 99: The test is good, but splitting it into multiple would allow granularity in the output if something goes wrong

sz-piotr (changes requested):

sz-piotr:
@sin6pi7 I guess an alternative to adding `| 'all'` to the types would be removing the parameter in those cases

PR #149, opened 2022-05-27
Add on chain data encoding

sz-piotr:


packages/encoding/src/encoding/writePositionUpdate.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export function writePositionUpdate(
+  writer: ByteWriter,
+  position: PositionUpdate
+) {
+  writer.writeNumber(position.balances.length + 4, 32)
+  writer.writeNumber(position.positionId, 32)
+  writer.write(position.publicKey.toString(), 32)
+  writer.writeNumber(position.collateralBalance - MIN_INT, 32)
+  writer.writeNumber(Number(position.fundingTimestamp) / 1000, 32)
```
sin6pi7, line 14: You could use `toSeconds` here.

sin6pi7 (approved):

PR #151, opened 2022-05-27
Test network

sz-piotr:


packages/frontend/src/scripts/metamask.ts
```diff
@@ -50,7 +50,7 @@ export function initMetamask() {
   }
 
   function updateChainId(chainId: string) {
-    if (chainId !== '0x1') {
+    if (chainId !== '0x1' && chainId !== '0x539') {
```
sin6pi7, line 53: If these were assigned to constants it would be more readable what chains these are.

sin6pi7 (approved):

PR #162, opened 2022-05-31
Display offers on position page BE logic

michalsidzej:


packages/backend/src/peripherals/database/ForcedTradeOfferRepository.ts
```diff
@@ -188,6 +188,13 @@ export class ForcedTradeOfferRepository extends BaseRepository {
     return rowIds.map(String).map(AssetId)
   }
 
+  async getByPositionId(position_id_a: bigint) {
+    const rows = await this.knex('forced_trade_offers')
+      .where({ position_id_a })
```
sin6pi7, line 193:
```suggestion
  async getPendingByPositionIdA(positionIdA: bigint) {
    const rows = await this.knex('forced_trade_offers')
      .where({ position_id_a: positionIdA })
```

1. Lets keep variable naming consistent (we use camelCase),
2. The query returns data specifically for `positionIdA`, so the signature should reflect that - otherwise it could mean it returns also offers for `positionIdB`,
3. The query return only pending offers (i.e. initial + accepted).

packages/backend/src/peripherals/database/ForcedTradeOfferRepository.ts
```diff
@@ -188,6 +188,13 @@ export class ForcedTradeOfferRepository extends BaseRepository {
...
+      .whereNull('transaction_hash')
```
sin6pi7, line 194:
```suggestion
      .whereNull('transaction_hash')
      .whereNull('cancelled_at')
```

We don't want to show cancelled offers.

PR #167, opened 2022-06-01
Fix displaying ownedByYou on position details page

michalsidzej:


packages/frontend/src/pages/positions/PositionDetails.tsx
```diff
@@ -142,7 +142,7 @@ export function PositionDetails({
   account,
   pendingOffers,
 }: PositionDetailsProps) {
-  const ownedByYou = ethAddress === account
+  const ownedByYou = !!account && ethAddress === account.address
```
sin6pi7, line 145:
```suggestion
  const ownedByYou = account?.address === ethAddress
```

Wouldn't this work as expected?

sin6pi7 (approved):

PR #170, opened 2022-06-02
Display forced txs count on state update list

michalsidzej:


packages/frontend/src/pages/home/Home.tsx
```diff
@@ -65,6 +66,9 @@ export function Home(props: HomeProps) {
               formatHashLong(update.hash),
               formatRelativeTime(update.timestamp),
               update.positionCount.toString(),
+              update.forcedTxsCount > 0
+                ? update.forcedTxsCount.toString()
+                : '-',
```
sz-piotr, line 71: It's absolutely fine to show 0

sz-piotr (changes requested):

packages/frontend/src/pages/home/Home.tsx
michalsidzej: It's just the design
Screenshot 2022-06-03 at 10 01 31

sz-piotr (approved):

PR #172, opened 2022-06-02
Redirect offer

sin6pi7:
Requires https://github.com/l2beat/dydx-state-explorer/pull/171 to be merged first.


packages/backend/src/api/controllers/ForcedTradeOfferController.ts
```diff
@@ -85,6 +85,13 @@ export class ForcedTradeOfferController {
       }
     }
 
+    if (offer.accepted?.transactionHash) {
+      return {
+        type: 'redirect',
+        url: `/forced/${offer.accepted?.transactionHash}`,
```
sz-piotr, line 91: the `?` should no longer be needed here

sz-piotr (changes requested):

sz-piotr (approved):

PR #177, opened 2022-06-03
Unify forced action naming

sin6pi7:


packages/testnet/contracts/Perpetual.sol
```diff
@@ -8,20 +8,20 @@ contract Perpetual {
 
     event LogForcedWithdrawalRequest(
```
sz-piotr, line 9: The solidity source code is the one place I would actually keep the original naming, because this will allow us to make sure that it works with those names.

sz-piotr (changes requested):

packages/testnet/contracts/Perpetual.sol
sin6pi7: Makes sense.

sz-piotr (approved):

PR #200, opened 2022-06-23
Get logs by hash for blocks close to tip

sz-piotr:


packages/backend/src/peripherals/ethereum/EthereumClient.ts
```diff
@@ -36,11 +36,33 @@ export class EthereumClient {
     if (blockRange.isEmpty()) {
       return []
     }
-    const logs = await this.provider.getLogs({
-      ...filter,
-      fromBlock: blockRange.start,
-      toBlock: blockRange.end - 1,
-    })
+    let [from, to, hashes] = blockRange.splitByKnownHashes()
+    if (hashes.length > 40) {
```
krzkaczor, line 40: should 40 be SAFE_BLOCK_DISTANCE ?

packages/backend/src/peripherals/ethereum/EthereumClient.ts
```diff
@@ -36,11 +36,33 @@ export class EthereumClient {
...
+
+    const nestedLogs = await Promise.all([
+      from !== to
+        ? this.provider.getLogs({
+            ...filter,
+            fromBlock: from,
+            toBlock: to - 1,
+          })
+        : [],
+      ...hashes.map((blockHash) =>
```
krzkaczor, line 58: Have you tested it in practice? I think we might need to throttle concurrent calls here 🤔

sz-piotr: I have but I implemented batching anyway

packages/backend/src/peripherals/ethereum/EthereumClient.ts
sz-piotr: Moved the value to config

PR #204, opened 2022-07-26
Finalize forced exit

michalsidzej:


packages/backend/test/peripherals/database/ForcedTransactionsRepository.test.ts
```diff
@@ -239,7 +240,9 @@ describe(ForcedTransactionsRepository.name, () => {
     )
 
     const transactions = await repository.getByPositionId(positionId)
-    expect(transactions).toEqual([
+
+    expect(transactions).toBeAnArrayOfLength(2)
```
krzkaczor, line 244: why not use toEqual? you first check if array is of length 2 and later you assert that it has two exact elements. Wouldn't simple toEqual do the same job?

krzkaczor (commented):
1 small comm

packages/backend/test/peripherals/database/ForcedTransactionsRepository.test.ts
michalsidzej: We don't sort the elements here, while I was testing it the array had different order and caused the test to fail

krzkaczor: okay, but you can always sort the just before assertion. I think this would result in a more readable assertion, but it's a nit.

michalsidzej: Actually, when I think about it rn, we should sort the transactions by lastUpdate, I'll fix it now

krzkaczor (approved):

PR #207, opened 2022-07-29
Fix tracking finalize transaction

michalsidzej:


packages/backend/src/core/FinalizeExitEventsCollector.ts
```diff
@@ -0,0 +1,122 @@
...
+    this.getMinedTransactions =
+      _getMinedTransactions || this.getMinedTransactions
+  }
+
+  async collect(
+    blockRange: BlockRange
+  ): Promise<{ added: number; updated: number; ignored: number }> {
+    const transactions = await this.getMinedTransactions(blockRange)
+    const results = await Promise.all(
+      transactions.map(async ({ hash, data, minedAt, blockNumber }) => {
```
krzkaczor, line 54: two nits to increase readability of this function:
1. insert some new lines
2. short circuit to avoid nesting, ie: `if(transaction) return 'ignored'`

packages/backend/src/core/FinalizeExitEventsCollector.ts
```diff
@@ -0,0 +1,122 @@
...
+        const transaction =
+          await this.forcedTransactionsRepository.findByFinalizeHash(hash)
+        if (!transaction) {
+          const exitTransaction =
+            await this.forcedTransactionsRepository.findByFinalizeData(data)
+          if (!exitTransaction) {
+            return 'ignored'
+          }
+          this.forcedTransactionsRepository.saveFinalize(
```
krzkaczor, line 63: isn't this async fn? I think it should be awaited but what's more important, linter settings should be fixed to find these kind of errors automatically. Can you enable no-floating-promises rule?

krzkaczor (changes requested):

packages/backend/src/core/FinalizeExitEventsCollector.ts
michalsidzej: I'll add it as a separate PR, nice spot

packages/backend/src/core/FinalizeExitEventsCollector.ts
```diff
@@ -0,0 +1,244 @@
...
+  private syncExecuted = false
+  async oneTimeSync() {
+    if (this.syncExecuted) {
+      return
+    }
+
+    const firstSyncedBlock = await this.syncStatusRepository.getLastSynced()
+    if (!firstSyncedBlock || firstSyncedBlock < 14878490) {
+      return
+    }
```
sz-piotr, line 137: Doesn't this mean that we would run the oneTimeSync every time the server restarts? We should remember that we already done it by saving this in the database.

packages/backend/src/peripherals/database/ForcedTransactionsRepository.ts
```diff
@@ -365,15 +425,18 @@ export class ForcedTransactionsRepository extends BaseRepository {
...
+          hash: finalizeHash.toString(),
+          mined_at:
+            minedAt !== null && minedAt !== undefined
+              ? BigInt(minedAt.toString())
+              : null,
+          sent_at: sentAt !== null ? BigInt(sentAt.toString()) : null,
+          block_number: blockNumber,
+        })
+        .onConflict()
+        .ignore() // TODO: not sure
```
sz-piotr, line 439: What's up with this?

packages/backend/src/core/FinalizeExitEventsCollector.ts
michalsidzej: I thought about this, but imo we can just delete this part after it syncs. Thats the purpose of all those comments.

packages/backend/src/peripherals/database/ForcedTransactionsRepository.ts
michalsidzej: We prefer to ignore if conflicted on primary key, because there's a possibility we could override additional information (sent_at).

sz-piotr (approved):

PR #209, opened 2022-08-03
Remove floating promises

michalsidzej:


packages/backend/src/peripherals/database/ForcedTradeOfferRepository.ts
```diff
@@ -191,8 +191,8 @@ export class ForcedTradeOfferRepository extends BaseRepository {
   }
 
   private getByPositionIdQuery(positionId: bigint) {
-    return this.knex('forced_trade_offers').where(function () {
-      this.where({ position_id_a: positionId }).orWhere({
+    return this.knex('forced_trade_offers').where(async function () {
+      await this.where({ position_id_a: positionId }).orWhere({
         position_id_b: positionId,
       })
     })
```
sz-piotr, line 198: Does this work? I find it weird that it would, given that awaiting knex promises leads to query execution

packages/backend/src/peripherals/database/ForcedTransactionsRepository.ts
```diff
@@ -229,8 +229,8 @@ export class ForcedTransactionsRepository extends BaseRepository {
 
   async countPendingByPositionId(positionId: bigint) {
     const [{ count }] = await this.joinQuery()
-      .where(function () {
-        this.whereRaw("data->>'positionId' = ?", String(positionId))
+      .where(async function () {
+        await this.whereRaw("data->>'positionId' = ?", String(positionId))
```
sz-piotr, line 233: Same concern here

packages/backend/src/peripherals/database/ForcedTradeOfferRepository.ts
michalsidzej: After double checking it does in fact throw an error while running the server. There has to be a problem with our test setup, as the test run just fine with those awaits.

sz-piotr (approved):

PR #236, opened 2022-12-08
Create ValidiumSpotDataSyncService

Wendrowiec13:


packages/backend/src/core/ValidiumSpotDataSyncService.ts
```diff
@@ -0,0 +1,173 @@
...
+      )
+      const newPositions = this.buildNewPositions(batch)
+
+      await this.stateUpdater.processStateTransition(
+        {
+          id: id + i + 1,
+          blockNumber: transition.blockNumber,
+          stateTransitionHash: transition.stateTransitionFact,
+        },
+        dexOutput,
```
sz-piotr, line 86: You are still getting a typescript error here:
```
src/core/ValidiumSpotDataSyncService.ts(86,9): error TS2345: Argument of type 'StarkExDexOutput' is not assignable to parameter of type 'StarkExProgramOutput'.
 |   Type 'StarkExDexOutput' is missing the following properties from type 'StarkExProgramOutput': configurationHash, assetConfigHashes, oldState, newState, and 4 more.
```

sz-piotr (changes requested):

packages/backend/src/core/ValidiumSpotDataSyncService.ts
Wendrowiec13: That's right. I just spoke to Adrian about this and I will create a PR with just the decodeDexOutput changes, without calling it here

PR #237, opened 2022-12-12
Add VaultLeaf and rename Position to PositionLeaf

adamiak:
Resolves L2B-546


packages/state/src/RollupState.ts
```diff
@@ -21,7 +21,7 @@ export type FundingByTimestamp = Map<Timestamp, ReadonlyMap<AssetId, bigint>>
 export class RollupState {
   constructor(
     private readonly storage: IRollupStateStorage,
-    public readonly positions: MerkleTree<Position>
+    public readonly positionLeaves: MerkleTree<PositionLeaf>
```
sz-piotr, line 24: Maybe `positionTree` instead of `positionLeaves`?

packages/backend/test/peripherals/database/RollupStateRepository.test.ts
```diff
@@ -26,15 +26,15 @@ describe(RollupStateRepository.name, () => {
     })
 
     it('persists and recover a Position', async () => {
```
sz-piotr, line 28: The names inside strings should probably also be renamed.

packages/state/test/VaultLeaf.test.ts
```diff
@@ -0,0 +1,47 @@
+import { PedersenHash, StarkKey } from '@explorer/types'
+import { expect } from 'earljs'
+
+import { VaultLeaf } from '../src/VaultLeaf'
+
+const tokenA = PedersenHash(
+  '0xd5b742d29ab21fdb06ac5c7c460550131c0b30cbc4c911985174c0ea4a92ec'
+)
+
+describe(VaultLeaf.name, () => {
```
sz-piotr, line 10: Are the hash values in this test verified against starkware's code? When writing tests for Position I used their python resources.

For this you could use: https://github.com/starkware-libs/starkex-resources/blob/0f08e6c55ad88c93499f71f2af4a2e7ae0185cdf/stark_ex_objects/starkware/objects/state.py#L76

Alternatively, if you are getting correct state hashes with this implementation having the values be derived from what the code returns should also be fine.

sz-piotr (dismissed):

packages/state/test/VaultLeaf.test.ts
adamiak: Good resource! I checked hashes against hashes from DAC and it's generating correct ones. But the python implementation showed I should pack balance to 32 bytes, not 8 (which means my implementation would fail if balance wouldn't fit in 8 bytes). Fixed.

PR #238, opened 2022-12-13
Decode dex output

Wendrowiec13:
Resolves L2B-538


packages/encoding/src/decoding/decodeDexOutput.ts
```diff
@@ -0,0 +1,50 @@
...
+  const reader = new ByteReader(data)
+
+  const configCode = reader.readBigInt(32)
+  const initialValidiumVaultRoot = PedersenHash(reader.readHex(32))
+  const finalValidiumVaultRoot = PedersenHash(reader.readHex(32))
+  const initialRollupVaultRoot = PedersenHash(reader.readHex(32))
+  const finalRollupVaultRoot = PedersenHash(reader.readHex(32))
+  const initialOrderRoot = PedersenHash(reader.readHex(32))
+  const finalOrderRoot = PedersenHash(reader.readHex(32))
+  const globalExpirationTimestamp = reader.readBigInt(32)
```
sz-piotr, line 16: Probably it should be `readNumber` unless we're using `bigint` for timestamps everywhere else.

packages/encoding/src/decoding/decodeDexOutput.ts
```diff
@@ -0,0 +1,50 @@
...
+  const validiumVaultTreeHeight = reader.readNumber(32)
+  const rollupVaultTreeHeight = reader.readNumber(32)
+  const orderTreeHeight = reader.readNumber(32)
+  const nModifications = reader.readNumber(32)
+  const nConditionalTransfers = reader.readNumber(32)
+  const nL1VaultUpdates = reader.readNumber(32)
+  const nL1OrderMessages = reader.readNumber(32)
```
sz-piotr, line 23: instead of `nName` I'd use `nameCount` similarily to https://github.com/l2beat/starkex-explorer/blob/1525003d800f27596dd31ab81e44f79b778295da/packages/encoding/src/decoding/readFundingEntries.ts#L1

Also, what's the reason for `L1` prefix? maybe it can be dropped?

packages/encoding/src/OnChainData.ts
```diff
@@ -18,6 +18,25 @@ export interface StarkExProgramOutput {
   conditions: PedersenHash[]
 }
 
+//
```
sz-piotr, line 21: The comment should probably not be empty :)

sz-piotr (changes requested):

packages/encoding/src/decoding/decodeDexOutput.ts
Wendrowiec13: `minimumExpirationTimestamp` in `decodeFirstPage` is also read as `bigint` but `timestamp` in `decodeState` is read as `number`

packages/encoding/src/decoding/decodeDexOutput.ts
Wendrowiec13: The `L1` prefix comes straight from the Cairo code https://github.com/starkware-libs/starkex-for-spot-trading/blob/607f0b4ce507e1d95cd018d206a2797f6ba4aab4/src/starkware/cairo/dex/main.cairo#L21-L37

packages/encoding/src/OnChainData.ts
Wendrowiec13: Yup, fixed it

packages/encoding/src/decoding/decodeDexOutput.ts
sz-piotr: I see. In that case I'd use number here and in the future also modify minimumExpirationTimestamp to be a number too

packages/encoding/src/decoding/decodeDexOutput.ts
sz-piotr: @adamiak wdyt we should do about the naming here?

packages/encoding/src/decoding/decodeDexOutput.ts
Wendrowiec13: DOne

packages/encoding/src/decoding/decodeDexOutput.ts
```diff
@@ -0,0 +1,50 @@
...
+  const orderTreeHeight = reader.readNumber(32)
+  const modificationsCount = reader.readNumber(32)
+  const conditionalTransfersCount = reader.readNumber(32)
+  const L1VaultUpdatesCount = reader.readNumber(32)
+  const L1OrderMessagesCount = reader.readNumber(32)
+
+  if (!reader.isAtEnd()) {
+    // https://github.com/starkware-libs/starkex-contracts/blob/75c3a2a8dfff70604d851fc6b1a2bc8bc1a3964b/scalable-dex/contracts/src/components/OnchainDataFactTreeEncoder.sol#L12
+    // When reading calldata from updateState two new values are appended
+    reader.skip(64)
```
adamiak, line 28: I think we should remove this whole `.skip()` conditional block. If I remember correctly, it was added because ProgramOutput on Validium has 2 more fields that we didn't understand, so we decided to temporarily skip them. But here we assume we know the whole structure so we can `.assertEnd()`.

packages/encoding/src/decoding/decodeDexOutput.ts
adamiak: Switching from `nX` to `XCount` is 👍 

As for the `L1*` fields, I'm not sure what they're for. So for the time being I would keep the names (but with lowercase 'l', i.e. `l1VaultUpdateCount` etc.)

@Wendrowiec13 when adding `...Count` please make the word root singular, e.g. it should be `modificationCount` and not `modificationsCount`. From what I checked (on the web and with @sz-piotr) this would be more grammatically correct.

Wendrowiec13: Done

packages/encoding/src/decoding/decodeDexOutput.ts
Wendrowiec13: Done

adamiak (approved):

PR #239, opened 2022-12-13
Refactor rollup validium logic

adamiak:
This PR introduces no new logic at all. It only does refactoring and renaming. It also doesn't add new tests yet, but all existing ones pass.

What was refactored:
- DataSyncServices
- StateUpdaters
- Collectors

Regarding state updaters, now we have:
- `StateUpdater` with common logic
- `PerpetualRollupUpdater extends StateUpdater`
- `PerpetualValidiumUpdater extends StateUpdater`

Regarding data sync services, now we have:
- `PerpetualRollupSyncService` (renamed from `DataSyncService`)
- `PerpetualValidiumSyncService` 

Regarding collectors, now we have:
- `PerpetualRollupStateTransitionCollector` (renamed from `StateTransitionCollector`)
- `PerpetualValidiumStateTransitionCollector`
- `ProgramOutputCollector`

This structure makes it obvious where the upcoming logic for `SpotValidium...` sync service, state updater and collectors should go.

Side note: I'm aware that it may seem like there should be just `PerpetualStateUpdater` (regardless if it's a rollup or validium) but that's not the case. Updater takes data collected by collectors and it's different for both modes. In the future we can refactor the architecture into e.g. `Collectors -> *DataProcessor* -> Updater`, although due to the fact how rollup data is structured it might be more complex than it seems.


adamiak:
Closed as #240 is built on top of this branch.

PR #244, opened 2023-01-12
Add tests for Validium state transition collector

Wendrowiec13:


packages/backend/src/core/collectors/ValidiumStateTransitionCollector.test.ts
```diff
@@ -0,0 +1,216 @@
...
+        mock<EthereumClient>(),
+        stateTransitionRepository,
+        EthereumAddress.fake(),
+        LogRootUpdate
+      )
+
+      await collector.discardAfter(123)
+
+      expect(stateTransitionRepository.deleteAfter).toHaveBeenCalledWith([123])
+    })
```
Wendrowiec13, line 29: I don't know if the same test with `LogUpdateState` instead of `LogRootUpdate` would be useful

Wendrowiec13:
In the 2 tests for the `collect` we are basically doing the same thing. I wonder if there's a clever way to make it look cleaner

packages/backend/src/core/collectors/ValidiumStateTransitionCollector.test.ts
sz-piotr: I think it's fine to have a single test

sz-piotr (approved):

adamiak (approved):

PR #245, opened 2023-01-16
Add ForcedWithdrawRepository

sz-piotr:
Resolves L2B-622, L2B-638

This PR adds the following:
- new database tables. There are two tables for transaction type: one with data, the other with tx status. This is different than previously discussed where tx status was shared between all transactions. The reason is that withdraw txs have a different looking mined status, because once they are mined we know how much money was withdrawn.
- types for the new tables
- a test for migration numbers backported from the L2BEAT repo
- a small addition to the Timestamp type
- a repository for handling forced withdraw transactions

We'll soon see if having different tables for statuses per transaction type makes sense, because in a future PR I'll need to implement querying from both forced withdraw and forced trade txs to show this view https://dydx.l2beat.com/forced.

The main intent is of course for the changes to replace https://github.com/l2beat/starkex-explorer/blob/830f7c7f14b26a266c2992e64454eea3065dcc94/packages/backend/src/peripherals/database/ForcedTransactionRepository.ts as this is a nightmare to maintain.

### DIFF
--- a/packages/backend/src/peripherals/database/ForcedWithdrawRepository.ts
+++ b/packages/backend/src/peripherals/database/ForcedWithdrawRepository.ts
@@ -1,5 +1,8 @@
 import { Hash256, StarkKey, Timestamp } from '@explorer/types'
-import { ForcedWithdrawTransactionRow } from 'knex/types/tables'
+import {
+  ForcedWithdrawStatusRow,
+  ForcedWithdrawTransactionRow,
+} from 'knex/types/tables'
 
 import { Logger } from '../../tools/Logger'
 import { BaseRepository } from './shared/BaseRepository'
@@ -12,6 +15,47 @@ export interface ForcedWithdrawTransactionRecord {
   positionId: bigint
 }
 
+export interface ForcedWithdrawTransactionWithHistory
+  extends ForcedWithdrawTransactionRecord {
+  history: ForcedWithdrawStatus[]
+}
+
+export type ForcedWithdrawStatus =
+  | SentStatus
+  | MinedStatus
+  | RevertedStatus
+  | ForgottenStatus
+  | IncludedStatus
+
+export interface SentStatus {
+  status: 'sent'
+  timestamp: Timestamp
+}
+
+export interface MinedStatus {
+  status: 'mined'
+  timestamp: Timestamp
+  blockNumber: number
+}
+
+export interface RevertedStatus {
+  status: 'reverted'
+  timestamp: Timestamp
+  blockNumber: number
+}
+
+export interface ForgottenStatus {
+  status: 'forgotten'
+  timestamp: Timestamp
+}
+
+export interface IncludedStatus {
+  status: 'included'
+  timestamp: Timestamp
+  blockNumber: number
+  stateUpdateId: number
+}
+
 export class ForcedWithdrawRepository extends BaseRepository {
   constructor(database: Database, logger: Logger) {
     super(database, logger)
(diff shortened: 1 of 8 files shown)


Wendrowiec13:
Lots of code to review. Need a little more time

packages/backend/src/peripherals/database/ForcedWithdrawRepository.ts
```diff
@@ -0,0 +1,368 @@
...
+  extends ForcedWithdrawTransactionRecord {
+  history: ForcedWithdrawStatus[]
+}
+
+export type ForcedWithdrawStatus =
+  | SentStatus
+  | MinedStatus
+  | RevertedStatus
+  | ForgottenStatus
+  | IncludedStatus
```
Wendrowiec13, line 29: Wouldn't it be better to declare those Statuses before using them? In terms of clean code of course. I don't know if we have a "rule" for this

packages/backend/src/peripherals/database/ForcedWithdrawRepository.ts
```diff
@@ -0,0 +1,368 @@
...
+    timestamp: Timestamp
+  }): Promise<Hash256> {
+    const knex = await this.knex()
+    await knex('forced_withdraw_statuses').insert({
+      hash: status.hash.toString(),
+      status: 'forgotten',
+      timestamp: BigInt(status.timestamp.toString()),
+    })
+    return status.hash
+  }
```
Wendrowiec13, line 149: What are forgotten transactions?

packages/backend/src/peripherals/database/ForcedWithdrawRepository.ts
```diff
@@ -0,0 +1,368 @@
...
+    const knex = await this.knex()
+    await knex('forced_withdraw_statuses').insert({
+      hash: status.hash.toString(),
+      status: 'included',
+      timestamp: BigInt(status.timestamp.toString()),
+      block_number: status.blockNumber,
+      state_update_id: status.stateUpdateId,
+    })
+    return status.hash
+  }
```
Wendrowiec13, line 166: What's the difference between included and mined transactions?

packages/backend/src/peripherals/database/shared/types.ts
```diff
@@ -155,6 +153,59 @@ declare module 'knex/types/tables' {
...
+  interface WithdrawStatusRow {
+    hash: string
+    status: string
+    timestamp: bigint
+    block_number: Nullable<number>
+    quantized_amount: Nullable<bigint>
+    non_quantized_amount: Nullable<bigint>
+    recipient_address: Nullable<string>
+  }
+
```
Wendrowiec13, line 208: Do you think it's better to have Statuses types in `ForcedWithdrawRepository` instead of here?

sz-piotr: The reason the statuses (but only names) are also declared in the types is to make sure that the `switch` at https://github.com/l2beat/starkex-explorer/pull/245/files#diff-5cac483fadd9b48a926df5ff307622a120daac55611bcf9f5286ed9f172f3f1aR324 is exhaustive.

packages/backend/src/peripherals/database/ForcedWithdrawRepository.ts
sz-piotr: Mined means that the transaction appeared on the blockchain. Included means that a state update referenced that particular forced transaction. That usually happens several hours later. It's only something relevant to forced transactions in StarkEx

packages/backend/src/peripherals/database/ForcedWithdrawRepository.ts
sz-piotr: Sometimes if you send a transaction with a low fee and it doesn't get included in a long time or if you send a transaction that becomes replaced by another transaction (this can happen if you send two transactions with the same nonce and is usually used to increase the tx fee or "cancel" transactions) the nodes drop your transaction and it never ever becomes mined. We have this status to be able to account for this possibility

packages/backend/src/peripherals/database/ForcedWithdrawRepository.ts
Wendrowiec13: Ok, makes sense now, thanks

packages/backend/src/peripherals/database/ForcedWithdrawRepository.ts
Wendrowiec13: Makes sense I would make it a convention if that's what we do anyways

packages/backend/src/peripherals/database/shared/types.ts
Wendrowiec13: Ok, that brings me a different thing I now thought of. Would it be better to have an enum for the statuses names?

sz-piotr: Enums have a lot of problems in TypeScript, because they work differently than other language features. Most of what TypeScript does happens only on the type level and gets removed during compilation. Enums persist and stay in the code, which is unintuitive and makes them not play nicely with other typescript constructions that don't behave in this way. Enums are also hard to debug because they are essentially variables with primitive values (usually consecutive integers) assigned to them, so if you console log an enum you don't get any indication that this value is an enum, it will just be a regular primitive

Wendrowiec13 (approved):

PR #246, opened 2023-01-16
Decode last two fields of PerpetualCairoOutput if present

Wendrowiec13:


Wendrowiec13:
I theory we could now differentiate between PerpetualRollup and Perpetual Validium modes type wise but I don't think that's necessary since we won't be using these new fields anyway

sz-piotr (approved):

PR #252, opened 2023-01-23
Refactor transaction handling

sz-piotr:
Resolves L2B-623, L2B-624, L2B-637, L2B-639, L2B-640

This PR completely rewrites how transactions are handled on the backend while trying to not affect the frontend "too much". Some differences are unfortunately inevitable but acceptable.

Missing things: (I'll go over them soon, but they aren't blocking the review)
   - [x] tests for SentTransactionRepository
   - [x] test against production data once again (I already did this once but fixed bugs since) to make sure everything works
   - [x] test transaction sending against local testnet

Below is a guide for reviewing this PR:
1. SentTransactionRepository.ts
   - pay special attention to `add` and `findFirstWithdrawByStarkKeyAfter`
   - note the lack of `deleteAfter`, we persist sent transactions regardless of reorgs
   - note the presence of `deleteByTransactionHash`, we actually forget some transactions if needed
1. SentTransaction.ts
   - understand that the `Encoded` type is used to guide queries in `SentTransactionRepository`
   - note the `ToJSON` type and it making the code smaller and less repetetive
   - note that only `positionA` in forced trades is queryable in `SentTransactionRepository`
1. UserTransaction.ts
   - same concept as `SentTransaction`, but for mined transactions
   - note that the `Encoded` type here supports queries by `positionA` and `positionB`
1. UserTransactionRepository.ts
   - pay special attention to `add`, `addManyIncluded`, `getByStateUpdateIdAndPositionId`, `getPaginated`
   - note the optional `types` parameter in a lot of queries. It's very important for type safety and future proofing
   - note that the repository manages two tables: `user_transactions` and `included_forced_requests`
1. UserTransactionCollector.ts
   - This now queries for all different transaction types based on events and combines Forced* and regular Withdrawal txs
   - note `knownBlockTimestamps`. It is not necessary, but helps massively when doing the software migration (more later)
1. StateUpdater.processStateTransition
   - note how we now call `userTransactionRepository.addManyIncluded` there
1. TransactionSubmitController
1. TransactionStatusService
   - note that it now also schedules itself
1. PositionController
   - note how `getPositionDetailsPage` combines sent and mined transactions to display actual list
   - note that the timestamp of the transaction is when it was sent or when it was mined. Any other events like inclusion in a state update or subsequent withdrawals don't affect this timestamp!
1. ForcedTransactionController
   - note how it handles showing finalize transactions. They don't show up as a status on the list and only exist on the details view
   - note that sent transactions are also only relevant on the details view
1. StateUpdateController
1. UserTransactionMigrator
   - it is used to run a one time code path
   - note how its resilient to errors through the use of `softwareMigrationRepository.setMigrationNumber`
   - redownloads all user transactions, because we didn't track withdrawals in full before
   - migrates inclusion in state updates from existing data
   - migrates sent transaction partly through existing data, but redownloads specifics to fill gaps
1. Everything else, sorry :(


packages/backend/src/peripherals/database/ForcedTradeOfferRepository.ts
```diff
@@ -146,13 +147,24 @@ export class ForcedTradeOfferRepository extends BaseRepository {
...
+  }
+
+  async updateTransactionHash(
+    id: number,
+    transactionHash: Hash256
+  ): Promise<number> {
+    const knex = await this.knex()
+    const updates = await knex('forced_trade_offers')
+      .update({ transaction_hash: transactionHash.toString() })
+      .andWhere({ id })
```
sz-piotr, line 166: TODO: Test and replace `andWhere` with `where`

packages/backend/src/core/StateUpdater.ts
```diff
@@ -101,21 +109,48 @@ export class StateUpdater<T extends PositionLeaf | VaultLeaf> {
...
-      )
-    const filteredHashes = hashes.filter(
-      (h): h is Exclude<typeof h, undefined> => h !== undefined
-    )
+    const notIncluded = await this.userTransactionRepository.getNotIncluded([
+      'ForcedTrade',
+      'ForcedWithdrawal',
+    ])
+
+    return forcedActions.map((action) => {
```
sz-piotr, line 117: TODO: test this

packages/backend/src/peripherals/ethereum/EthereumClient.ts
```diff
@@ -79,13 +80,66 @@ export class EthereumClient {
...
+          }
+          return this.provider.getLogs(filter)
+        })
       )
       logs.push(...nestedLogs.flat())
     }
     return logs
   }
 
+  async getAllLogs(
```
adamiak, line 105: It would be useful to test this function, especially the binary search part.

adamiak (approved):

packages/backend/src/peripherals/ethereum/EthereumClient.ts
sz-piotr: ah yes. I've copied it straight from L2BEAT's repository, I'll just copy the tests too

PR #259, opened 2023-02-08
New User page

Wendrowiec13:
Resolves L2B-762
Resolves L2B-763
Resolves L2B-764
Resolves L2B-765
Resolves L2B-766
Resolves L2B-767
Resolves L2B-768


packages/frontend/src/pages/newUser/BalanceChangesTable.tsx
```diff
@@ -0,0 +1,59 @@
...
+              href={`/stateUpdates/${balanceChange.stateUpdateId}`}
+              className="text-blue-300 underline"
+            >
+              #{balanceChange.stateUpdateId}
+            </a>,
+            balanceChange.asset,
+            balanceChange.newBalance.toString(),
+            <p
+              className={cx(
+                '',
```
torztomasz, line 46: {positive change ? 'text-green-100' : 'text-red-100'}

packages/frontend/src/pages/newUser/AssetsTable.tsx
```diff
@@ -0,0 +1,45 @@
...
+export function AssetsTable({ assets }: AssetsTableProps) {
+  return (
+    <NewTable
+      fullBackground
+      pageSize={6}
+      id="test"
+      title="Assets"
+      noRowsText="You have no assets"
+      columns={[
+        { header: '', className: '!w-9', textAlignClass: 'text-right' },
```
torztomasz, line 20: extracting to variable would help with readability

packages/frontend/src/pages/common/Button.tsx
```diff
@@ -0,0 +1,31 @@
+import cx from 'classnames'
+import React from 'react'
+
+interface ButtonProps {
+  readonly onClick?: () => void
+  readonly className?: string
+  readonly children: React.ReactNode
+  readonly variant: 'ACTION' | 'VIEW_ALL'
```
torztomasz, line 8: I believe we should name it 'primary' | 'outline'

packages/frontend/src/pages/common/Button.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export function Button({ onClick, className, variant, children }: ButtonProps) {
```
torztomasz, line 11: variant = 'ACTION' by default

packages/frontend/src/pages/common/Button.tsx
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <button
+      onClick={onClick}
+      className={cx(
+        'cursor-pointer !m-0',
+        {
```
torztomasz, line 17: probably better to extract it to classnameMap: Record<ButtonVariant, string>

packages/frontend/src/pages/common/Button.tsx
```diff
@@ -0,0 +1,31 @@
+import cx from 'classnames'
+import React from 'react'
+
+interface ButtonProps {
```
torztomasz, line 4:
```
interface ButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
  readonly variant?: ButtonVariant
}
```

packages/frontend/src/pages/newUser/Status.tsx
```diff
@@ -0,0 +1,50 @@
...
+  'SENT (1/2)': 'bg-blue-400',
+  'MINED (2/2)': 'bg-green-500',
+  REVERTED: 'bg-red-300',
+}
+
+interface StatusProps {
+  status: OfferEntry['status'] | EthereumTransactionEntry['status']
+}
+
+export function Status({ status }: StatusProps) {
```
torztomasz: Status -> StatusBadge

torztomasz (changes requested):

packages/frontend/src/pages/newUser/AssetsTable.tsx
Wendrowiec13: What do you mean?

packages/frontend/src/pages/newUser/Status.tsx
```diff
@@ -0,0 +1,50 @@
+import cx from 'classnames'
+import React from 'react'
+
+import { EthereumTransactionEntry, OfferEntry } from './UserProps'
+
+const backgroundMap = {
```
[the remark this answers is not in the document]
Wendrowiec13: Type extraction is not needed for that, but good point

packages/frontend/src/pages/newUser/UserProps.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export interface EthereumTransactionEntry {
+  readonly timestamp: Timestamp
+  readonly hash: string
+  readonly asset: string
+  readonly assetIcon: string
+  readonly amount: bigint
+  readonly status:
+    | 'SENT (1/3)'
```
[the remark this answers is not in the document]
Wendrowiec13: How does that benefit us right now?

packages/frontend/src/pages/newUser/BalanceChangesTable.tsx
Wendrowiec13: I think we should set a convention for conditional styles

packages/frontend/src/pages/newUser/UserProps.ts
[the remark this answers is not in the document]
torztomasz: I would say readability would increase + if you need to import status then you would import only the status not the whole type and then using key to get the type

packages/frontend/src/pages/newUser/AssetsTable.tsx
torztomasz:
```tsx
export function AssetsTable({ assets }: AssetsTableProps) {
  const columns = [
        { header: '', className: '!w-9', textAlignClass: 'text-right' },
        { header: 'NAME' },
        { header: 'BALANCE/ID' },
        { header: 'VAULT' },
        {
          header: 'ACTION',
          textAlignClass: 'text-left',
          className: '!w-[130px]',
        },
      ]
  const rows = assets.map((asset) => {
        const link = `/assets/${asset.name}` //TODO: Construct a proper link
        return {
          link,
          cells: [
            asset.icon,
            asset.name,
            asset.balance.toString(),
            asset.vaultId,
            <Button variant="ACTION">{asset.action}</Button>,
          ],
        }
      })

  return (
    <NewTable
      fullBackground
      pageSize={6}
      id="test"
      title="Assets"
      noRowsText="You have no assets"
      columns={columns}
      rows={rows}
    />
  )
}
```

torztomasz: I think we should avoid nesting maps and some big arrays in component props

Wendrowiec13: I would agree if this was a complex component. It's not the case here. I don't think the readability is impaired.

packages/frontend/src/pages/newUser/UserProps.ts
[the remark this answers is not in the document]
Wendrowiec13: I would argue that extracting to type here actually decreases readability. Right now we do not need the status anywhere else so extracting and exporting it is prioritising form over function in my opinion.

torztomasz (changes requested):

packages/frontend/tailwind.config.js
```diff
@@ -10,36 +10,39 @@ module.exports = {
       screens: {
         wide: '900px',
       },
-      borderRadius: {
-        'action-button': '0.25rem',
-      },
     },
```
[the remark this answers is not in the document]
Wendrowiec13: Done it here

torztomasz (approved):

PR #263, opened 2023-02-12
Rename token to assetHash for spot assets

adamiak:
Resolves L2B-824

Up until now we have used the name `token` and assumed it's a `PedersenHash` when dealing with **spot** assets (for example in `VaultLeaf` and `vaults` table). After recent refactoring we have a proper `AssetHash` type for spot assets. We also know it's not a Pedersen hash but a sliced Keccak.

This PR renames `token` to `assetHash` and the type from `PedersenHash` to `AssetHash`. The name `token` is preserved only in zed parsing code as this is the name that comes from DAC.


Wendrowiec13:
Screenshot 2023-02-12 at 20 52 08
Should't the type of the token property here also be changed?

packages/types/src/AssetHash.ts
```diff
@@ -7,15 +7,21 @@ export interface AssetHash extends String {
 }
 
 export function AssetHash(value: string) {
-  if (!value.startsWith('0x')) {
-    value = '0x' + value
+  if (value.startsWith('0x')) {
+    value = value.slice(2)
```
Wendrowiec13, line 11: I'm not sure if this is a better approach. We are slicing the `0x` prefix if it's there and then we are adding it anyway.

packages/types/src/AssetHash.ts
```diff
@@ -7,15 +7,21 @@ export interface AssetHash extends String {
...
   }
-  if (!/^0x[a-f\d]{64}$/i.test(value)) {
-    throw new TypeError('Invalid AssetHash')
+  if (!/^[\da-fA-F]+$/.test(value)) {
+    throw new TypeError('AssetHash must be a hex string')
   }
+  if (value.length > 64) {
+    throw new TypeError('AssetHash too large')
```
Wendrowiec13, line 17: Have you tested this? I am not entirely sure if it's correct

packages/types/src/AssetHash.ts
adamiak: This is done to pad the length to 64. So that 0x000ab and 0xab end up being equal.

adamiak:
> Shouldn't `PedershenHash0x` here be renamed to AssetHash?

@Wendrowiec13 good spot, fixed.

packages/state/src/VaultLeaf.ts
```diff
@@ -22,8 +22,8 @@ export class VaultLeaf extends MerkleValue {
   // https://github.com/starkware-libs/starkex-resources/blob/master/stark_ex_objects/starkware/objects/state.py#L76
   async calculateHash() {
     const key_token_hash = await pedersen(
-      PedersenHash(this.starkKey.substring(2)),
-      this.token
+      PedersenHash(this.starkKey.toString()),
+      PedersenHash(this.assetHash.toString())
```
sz-piotr, line 26: This suggests that even though it's a keccak hash we need to check if it starts with `0x0` in the `AssetHash` constructor, otherwise it will fail to be converted to `PedersenHash`.

sz-piotr (changes requested):

packages/types/src/AssetHash.ts
Wendrowiec13: Ok, makes sense then

PR #265, opened 2023-02-13
New design tailwind config

torztomasz:


sz-piotr (changes requested):
Use modern browsers option for fonts

Wendrowiec13 (approved):

PR #266, opened 2023-02-14
Add page component

torztomasz:
Resolves L2B-799, L2B-800


packages/frontend/src/view/components/common/Button.tsx
```diff
@@ -0,0 +1,33 @@
...
+interface ButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+  readonly variant?: ButtonVariant
+}
+
+type ButtonVariant = 'primary' | 'outline'
+
+const classNameMap: Record<ButtonVariant, string> = {
+  primary: 'bg-brand text-white',
+  outline: 'bg-transparent border border-brand text-white hover:bg-opacity-20',
+}
```
sz-piotr, line 13: nitpick. I don't see a big benefit over inlining in `cx` `variant === 'primary' && '....'`

packages/frontend/src/view/components/common/icons/L2BeatIcon.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+
+export function L2BeatIcon(props: React.SVGProps<SVGSVGElement>) {
```
sz-piotr, line 3: I'd move `L2BeatIcon`, `L2BeatMinimalIcon`, `DydxIcon` and `StarkWareIcon` from `icons/` to `logos/` and rename to `L2BeatLogo`, `L2BeatLogoSquare`, `DydxLogoSquare` and `StarkWareLogo`.

This doesn't apply to things like `DiscordIcon` because even though this is a logo we treat it like a small monochrome icon.

packages/frontend/src/view/components/common/page/Favicons.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+
+export function Favicons() {
+  return (
+    <>
+      <link rel="shortcut icon" href="/favicon.png" />
+      <link rel="icon" href="/favicon.svg" />
+      <link rel="apple-touch-icon" href="/favicon.png" />
+      <link rel="mask-icon" href="/mask-icon.svg" color="#000000" />
```
sz-piotr, line 9: I'd comment out ones that we do not yet have static files for.

packages/frontend/src/view/components/common/page/Footer.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export function Footer() {
+  return (
+    <footer className="flex h-16 flex-wrap items-center	justify-between gap-y-3 whitespace-normal border-t border-t-zinc-800 p-6 text-sm">
+      <span className="flex">
+        Built by <L2BeatIcon height={18} width={44} className="mx-1" /> team and
+        funded by
+        <StarkWareIcon height={18} width={98} className="ml-1" />
+      </span>
+      <span className="text-zinc-500">@Copyright 2023 L2BEAT</span>
```
sz-piotr, line 14: why `@Copyright` instead of just `Copyright`? also `{new Date().getFullYear()}`

sz-piotr (changes requested):

packages/frontend/src/view/components/common/Button.tsx
torztomasz: I would say that if we add new ButtonVariant we would not get any typescript error that we have to add ```variant==='newVariant'``` to cx. With this approach you will get an error that you need to handle the new variant plus the code looks cleaner I would say

packages/frontend/src/view/components/common/page/Footer.tsx
torztomasz: That's actually something I wanted to talk about on some meeting
Once its like this: 
image
Once like this: 
image

torztomasz: Oh and now I see that it is not @ but ©.

packages/frontend/src/view/components/common/icons/L2BeatIcon.tsx
torztomasz: That's good catch. What do you think about creating new folder inside view called assets? Inside it I would create logos and icons

torztomasz: Cuz its neither component or page

packages/frontend/src/view/components/common/page/Footer.tsx
sz-piotr: Let's just drop the symbol and go with `Copyright`

torztomasz: sure

packages/frontend/src/view/components/common/page/Footer.tsx
```diff
@@ -0,0 +1,19 @@
...
+export function Footer() {
+  return (
+    <footer className="flex h-16 flex-wrap items-center	justify-between gap-y-3 whitespace-normal border-t border-t-zinc-800 p-6 text-sm">
+      <span className="flex">
+        Built by <L2BeatLogo height={18} width={44} className="mx-1" /> team and
+        funded by
+        <StarkWareLogo height={18} width={98} className="ml-1" />
+      </span>
+      <span className="text-zinc-500">
+        Copyright ${new Date().getFullYear()} L2BEAT
```
sz-piotr, line 15:
```suggestion
        Copyright {new Date().getFullYear()} L2BEAT
```

sz-piotr (changes requested):

PR #270, opened 2023-02-16
Implement new forced actions page

torztomasz:
Resolves L2B-833


Wendrowiec13:
What was the reason you put `AssetDetails` in the shared package instead of in the types package?

packages/frontend/src/preview/data.ts
```diff
@@ -67,3 +68,127 @@ export const USER_PROPS: UserPageProps = {
...
+  assets: [
+    {
+      assetId: AssetId('USDC-6'),
+      balance: 69420_654321n,
+      priceUSDCents: 100n,
+      totalUSDCents: 69420_65n,
+    },
+    {
+      assetId: AssetId('ETH-9'),
+      balance: 21_370000000n,
```
Wendrowiec13, line 161: 🚀

packages/frontend/src/utils/assetUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+function assetHashToInfo(
+  assetHash: AssetHash,
+  assetDetails?: AssetDetails
+): AssetInfo {
+  if (!assetDetails) {
+    return {
+      name: 'Unknown',
+      symbol: assetHash.toString(),
```
Wendrowiec13, line 45: It's not a good idea to use AssetHash converted to string as symbol here. I would use 'unknown' or something like that instead.

packages/frontend/src/utils/assetUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+    //eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const ethInfo = perpetualsInfo.ETH!
+    return {
+      ...ethInfo,
+      imageUrl: getDydxImageUrl(ethInfo.symbol),
+    }
+  }
+  return {
+    name: assetDetails.name ?? assetDetails.symbol ?? 'Unknown',
+    symbol: assetDetails.symbol ?? assetDetails.name ?? assetHash.toString(),
```
Wendrowiec13, line 60: Same as above

packages/frontend/src/utils/assetUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+
+function getTrustWalletImageUrl(address: string) {
+  return `https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/${address}/logo.png`
+}
+
+function getDydxImageUrl(symbol: string) {
+  return `https://trade.dydx.exchange/currencies/${symbol.toLowerCase()}.svg`
+}
+
+const perpetualsInfo: Record<string, Omit<AssetInfo, 'imageUrl'>> = {
```
Wendrowiec13, line 74: I think we should move it into a separate file

packages/frontend/src/view/components/common/Button.tsx
```diff
@@ -1,33 +1,33 @@
...
   children,
+  disabled,
   ...rest
 }: ButtonProps) {
   return (
     <button
-      onClick={onClick}
       className={cx(
-        '!m-0 cursor-pointer font-semibold text-white',
+        'mx-auto min-w-[150px] rounded-lg px-10 py-2.5 text-sm font-semibold text-white disabled:cursor-not-allowed',
```
Wendrowiec13, line 26: Contained and outlined variants have different borderRadius, textSize and padding.

packages/frontend/src/view/components/common/FancyList.tsx
```diff
@@ -0,0 +1,41 @@
+import cx from 'classnames'
+import React from 'react'
+
+interface FancyListProps {
+  readonly items: React.ReactNode[]
+  readonly className?: string
+}
+
+export function FancyList({ items, className }: FancyListProps) {
```
Wendrowiec13, line 9: FancyList is not a good name for a component

packages/frontend/src/view/pages/forcedactions/ForcedWithdrawPage.tsx
```diff
@@ -0,0 +1,86 @@
...
+    <Page path="/forced" description="Description" account={props.account}>
+      <div className="my-auto flex gap-12">
+        <div className="flex flex-col">
+          <span className="text-2xl font-semibold">
+            Begin withdrawal process
+          </span>
+          <span className="mt-6 text-sm font-semibold text-zinc-500">
+            The withdrawal process consists of three steps:
+          </span>
+          <FancyList items={instructions} className="mt-3 max-w-[438px]" />
```
Wendrowiec13, line 45: Setting max width to an arbitrary value in px is not a good idea

packages/frontend/src/preview/data.ts
torztomasz: XD

packages/frontend/src/view/components/common/Button.tsx
torztomasz: Then I believe it is some mistake, they should only be differed by fill and outline. I will talk with Janek about it

packages/frontend/src/utils/assetUtils.ts
torztomasz: That is how it is designed
image

I think only thing missing is shortening of an address

packages/frontend/src/view/pages/forcedactions/ForcedWithdrawPage.tsx
```diff
@@ -0,0 +1,90 @@
...
+} from './ForcedActionFormProps'
+
+function ForcedWithdrawPage(props: ForcedActionFormProps) {
+  const instructions = [
+    <span>
+      Using this form you request a withdrawal of your funds. (Perpetual only)
+      You can only withdraw USDC, so to fully exit all funds you should first
+      get rid of your synthetic assets by using the close functionality.
+      (Perpetual only) This is achieved through a mechanism called forced
+      withdrawals, <Link href="https://google.com">link to docs</Link>.
```
sz-piotr, line 24: Don't use random urls. If you don't have a link 1) ask 2) if there is none use `#`. In this case the link is `https://docs.starkware.co/starkex/perpetual/perpetual-trading-forced-withdrawal-and-forced-trade.html#forced_withdrawal`

packages/frontend/src/view/pages/forcedactions/ForcedWithdrawPage.tsx
```diff
@@ -0,0 +1,90 @@
...
+            id={FormId.Form}
+            className="flex flex-col gap-6"
+            data-props={propsJson}
+          >
+            <div className="flex items-end justify-between">
+              <span className="text-2xl font-semibold">Withdrawal</span>
+              <span>
+                <span className="text-sm text-zinc-500">Position</span>{' '}
+                <span className="font-semibold">
+                  {props.positionId.toString()}
```
sz-piotr, line 62: All ids in the system (position, vault, state update, offer, etc) should be prefixed with `#` when displayed

packages/frontend/src/view/pages/forcedactions/ForcedTradePage.tsx
```diff
@@ -0,0 +1,100 @@
...
+function ForcedTradePage(props: ForcedActionFormProps) {
+  const instructions = [
+    'You create a trade offer using this form. This is fully off-chain and does not require any gas fees.',
+    'The trade offer will now be visible to all users of the system. You should seek out another user to accept the offer and become a counterparty to the trade. Accepting the offer is also fully off-chain and does not require any gas fees.',
+    <span>
+      <span className="text-yellow-300">
+        Once the offer is accepted you can submit a trade request.
+      </span>{' '}
+      This is achieved through a mechanism called forced trades,{' '}
+      <Link href="https://google.com">link to docs</Link>.
```
sz-piotr, line 29: https://docs.starkware.co/starkex/perpetual/perpetual-trading-forced-withdrawal-and-forced-trade.html#forced_trade

packages/frontend/src/view/pages/forcedactions/ForcedTradePage.tsx
```diff
@@ -0,0 +1,100 @@
...
+            className="flex flex-col gap-6"
+            data-props={propsJson}
+          >
+            <div className="flex items-end justify-between">
+              <span className="text-2xl font-semibold">
+                Forced {label} offer
+              </span>
+              <span>
+                <span className="text-sm text-zinc-500">Position</span>{' '}
+                <span className="font-semibold">11273</span>
```
sz-piotr, line 65: Should use position id from props (with a `#` symbol)

packages/frontend/src/view/pages/forcedactions/ForcedTradePage.tsx
```diff
@@ -0,0 +1,100 @@
...
+              </div>
+              <ForcedActionCard>
+                <TotalInput />
+              </ForcedActionCard>
+            </div>
+            <div className="flex flex-col gap-2">
+              <Button className="w-full" id={FormId.SubmitButton}>
+                Create {label} offer
+              </Button>
+              <Button className="w-full" variant="outlined">
```
sz-piotr, line 87: Should be a link to `/user/:starkKey`

packages/frontend/src/view/pages/forcedactions/ForcedWithdrawPage.tsx
```diff
@@ -0,0 +1,90 @@
...
+            <div className="flex flex-col gap-2">
+              <ForcedActionCard>
+                <AmountInput {...props} />
+              </ForcedActionCard>
+            </div>
+            <div className="flex flex-col gap-2">
+              <Button className="w-full" id={FormId.SubmitButton}>
+                Prepare for withdrawal
+              </Button>
+              <Link href="/forced">
```
sz-piotr, line 75: Should be a link to /user/:starkKey

packages/frontend/src/view/pages/forcedactions/ForcedWithdrawPage.tsx
```diff
@@ -0,0 +1,90 @@
...
+                <Button className="w-full" variant="outlined" type="button">
```
sz-piotr, line 76: Don't nest buttons inside links. Rather style the link like the button.

packages/frontend/src/view/components/common/OrderedList.tsx
```diff
@@ -0,0 +1,39 @@
...
+import React from 'react'
+
+interface FancyListProps {
+  readonly items: React.ReactNode[]
+  readonly className?: string
+}
+
+export function OrderedList({ items, className }: FancyListProps) {
+  return (
+    <div className={cx('ml-6', className)}>
```
sz-piotr, line 11: Why not ol + li?

sz-piotr (changes requested):

packages/frontend/src/view/pages/forcedactions/ForcedWithdrawPage.tsx
torztomasz: But then we would have to keep the same styles in Link component, maybe we can add href prop to Button?

sz-piotr (approved):

PR #278, opened 2023-02-21
Calculating withdrawable assets

adamiak:
Resolves L2B-801

This PR adds collection of events that:
- move assets to "withdrawal area" (pendingWithdrawals map in solidity)
- finalize the withdrawal

All these events create records in `withdrawable_assets` table with `balance_delta` column being negative or positive. By summing on this column we can check if there are any pending withdrawals for a user.

Logs that finalize the withdrawal also go to `user_transactions` table. Therefore these events are handled by `UserTransactionCollector` which send them both to `UserTransactionRepository` and `WithdrawableAssetsRepository`. 

Events moving assets to "withdrawal area" are collected by `WithdrawalAllowedCollector`


packages/backend/src/peripherals/database/transactions/UserTransactionRepository.ts
```diff
@@ -47,6 +48,10 @@ export type UserTransactionAddRecord = Parameters<
   UserTransactionRepository['add']
 >['0']
 
+export type WithdrawableAssetAddRecord = Parameters<
```
sz-piotr, line 51: I don't think this export belongs in this file

sz-piotr (changes requested):

packages/backend/src/peripherals/database/transactions/UserTransactionRepository.ts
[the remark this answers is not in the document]
adamiak: Moved

packages/backend/src/core/collectors/WithdrawableAssetCollector.ts
```diff
@@ -0,0 +1,169 @@
...
+        [
+          // Assets being moved to withdrawal area:
+          LogWithdrawalAllowed.topic,
+          LogMintableWithdrawalAllowed.topic,
+          LogAssetWithdrawalAllowed.topic,
+
+          // Assets being withdrawn from L2:
+          LogWithdrawalPerformed.topic,
+          LogWithdrawalWithTokenIdPerformed.topic,
+          LogMintWithdrawalPerformed.topic,
```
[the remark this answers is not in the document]
adamiak: Fixed. These events are now handled by `UserTransactionCollector`, which passes them to both `UserTransactionRepository` and `WithdrawableAssetRepository`  (with different encoding) . There is a separate `WithdrawalAllowedCollector` which collects only events that move assets to "withdrawal area" - these are not connected to any Ethereum transactions, so they are not going to `UserTransactionRepository`

packages/backend/src/core/collectors/UserTransactionCollector.ts
```diff
@@ -70,20 +89,8 @@ export class UserTransactionCollector {
       timestamp: Timestamp.fromSeconds(timestamp),
     }
 
+    let record
```
Wendrowiec13, line 92: We could type this here

packages/backend/src/peripherals/database/WithdrawableAssetRepository.test.ts
```diff
@@ -0,0 +1,227 @@
...
+
+      expect(record1?.transactionHash).toEqual(fake1.transactionHash)
+      expect(record1?.blockNumber).toEqual(fake1.blockNumber)
+      expect(record1?.timestamp).toEqual(fake1.timestamp)
+      expect(record1?.data).toEqual(fake1.data)
+      expect(record2?.data).toEqual(fake2.data)
+      expect(record3?.data).toEqual(fake3.data)
+      expect(record4?.data).toEqual(fake4.data)
+      expect(record5?.data).toEqual(fake5.data)
+      expect(record6?.data).toEqual(fake6.data)
```
Wendrowiec13, line 176: Is there a reason why we are not doing this in a loop?

Wendrowiec13 (approved):

packages/backend/src/peripherals/database/WithdrawableAssetRepository.test.ts
adamiak: I prefer to be more explicit in tests - adding a loop can introduce subtle bugs.

packages/backend/src/core/collectors/UserTransactionCollector.ts
adamiak: Given how simple update this is, I'll do it in a separate PR, because there are other PRs that rely on this one and I want to keep merging simple.

PR #285, opened 2023-02-22
Add UserController and StateUpdateController

adamiak:
Resolves L2B-804, L2B-808


packages/backend/src/peripherals/database/transactions/UserTransactionRepository.ts
```diff
@@ -62,6 +62,7 @@ export class UserTransactionRepository extends BaseRepository {
       this.getByStateUpdateIdAndPositionId
     )
     this.getPaginated = this.wrapGet(this.getPaginated)
+    this.getPaginatedByStarkKey = this.wrapGet(this.getPaginatedByStarkKey)
```
sz-piotr, line 65: getByStarkKeyPaginated

sz-piotr (approved):

packages/backend/src/peripherals/database/transactions/UserTransactionRepository.ts
adamiak: Fixed

PR #297, opened 2023-02-26
Display transactions lists

adamiak:
Resolves L2B-805

Please first review: https://github.com/l2beat/starkex-explorer/pull/278 - this is built on that branch.

Shows transactions list on Home Page,  UserPage, and paginated versions of both.


packages/backend/src/api/controllers/userTransactionToEntry.ts
```diff
@@ -0,0 +1,84 @@
...
+    case 'ForcedTrade':
+      return data.isABuyingSynthetic ? 'FORCED_BUY' : 'FORCED_SELL'
+    case 'FullWithdrawal':
+      return 'FORCED_WITHDRAW'
+    case 'Withdraw':
+      return 'WITHDRAW'
+    case 'WithdrawWithTokenId':
+      return 'WITHDRAW'
+    case 'MintWithdraw':
+      return 'WITHDRAW'
```
Wendrowiec13, line 66: We don't have to return `WITHDRAW` 3 times. We can have just one return here

packages/backend/src/peripherals/database/transactions/UserTransactionRepository.ts
```diff
@@ -123,9 +125,31 @@ export class UserTransactionRepository extends BaseRepository {
...
       .orderBy('timestamp', 'desc')
+
+    if (pagination) {
+      query = query.limit(pagination.limit).offset(pagination.offset)
+    }
+
     return toRecords<T>(await query)
   }
 
+  async getCountByStarkKey(
```
Wendrowiec13, line 136: I am not sure whether this function belongs here. Shouldn't repositories deal only with inserting, editing, deleting and retrieving data?

Wendrowiec13 (approved):

packages/backend/src/peripherals/database/transactions/UserTransactionRepository.ts
adamiak: Calculating count is also a query to the database, in this sense it's a "read". I think it's the right place.

packages/backend/src/api/controllers/userTransactionToEntry.ts
adamiak: Fixed

PR #308, opened 2023-02-28
Add gammax and myria logos

torztomasz:
Resolves L2B-947


adamiak (dismissed):

packages/frontend/src/view/components/page/Navbar.tsx
```diff
@@ -13,6 +15,7 @@ export interface NavbarProps {
 }
 
 export function Navbar({ user, searchBar = true }: NavbarProps) {
+  const starkExInstance = process.env.STARKEX_INSTANCE ?? 'dydx-mainnet'
```
adamiak, line 18: Don't use process.env on frontend. Everything will be pased as props

PR #312, opened 2023-02-28
Reimplement min and max due to stack limit

adamiak:


adamiak:
Fixed on different PR

PR #315, opened 2023-02-28
Add TransactionController

adamiak:
Resolves L2B-806, L2B-807

### DIFF
--- a/packages/backend/src/api/controllers/TransactionController.ts
+++ b/packages/backend/src/api/controllers/TransactionController.ts
@@ -1,0 +1,190 @@
+import {
+  PerpetualForcedWithdrawalPageProps,
+  renderOfferAndForcedTradePage,
+  renderPerpetualForcedWithdrawalPage,
+  renderSpotForcedWithdrawalPage,
+} from '@explorer/frontend'
+import { UserDetails } from '@explorer/shared'
+import { EthereumAddress, Hash256 } from '@explorer/types'
+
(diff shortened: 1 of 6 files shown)


packages/backend/src/api/controllers/TransactionController.ts
```diff
@@ -0,0 +1,254 @@
...
+      this.sentTransactionRepository.findByTransactionHash(txHash),
+      this.userTransactionRepository.findByTransactionHash(txHash),
+    ])
+
+    // TODO: cover the case where there is ONLY a sentTransaction
+    if (!userTransaction) {
+      return { type: 'not found', content: 'Transaction not found' }
+    }
+
+    let content
```
Wendrowiec13, line 48: We could probably type this here

packages/backend/src/api/controllers/TransactionController.ts
```diff
@@ -0,0 +1,254 @@
...
+  sentTransaction: SentTransactionRecord | undefined,
+  userTransaction: UserTransactionRecord
+): {
+  timestamp: Timestamp
+  status: 'SENT' | 'REVERTED' | 'MINED' | 'INCLUDED'
+}[] {
+  const history: {
+    timestamp: Timestamp
+    status: 'SENT' | 'REVERTED' | 'MINED' | 'INCLUDED'
+  }[] = []
```
Wendrowiec13, line 245: We could extract it to a type. It could extend the regular history type if we decided to extract that as well

PR #317, opened 2023-02-28
Allow singe starkKey to have multiple positions

adamiak:


packages/backend/src/core/preprocessing/SpotHistoryPreprocessor.ts
```diff
@@ -32,13 +32,19 @@ export class SpotHistoryPreprocessor extends HistoryPreprocessor<AssetHash> {
       if (vault.starkKey === StarkKey.ZERO) {
         await this.closePositionOrVault(trx, vault.vaultId, stateUpdate, {})
       } else {
-        const currentRecord =
-          await this.preprocessedAssetHistoryRepository.findCurrentByStarkKeyAndAsset(
-            vault.starkKey,
-            vault.assetHash,
+        const currentRecords =
+          await this.preprocessedAssetHistoryRepository.getCurrentByPositionOrVaultId(
+            vault.vaultId,
```
Wendrowiec13, line 37: Do we want to change find to get?

Wendrowiec13 (approved):

packages/backend/src/core/preprocessing/SpotHistoryPreprocessor.ts
adamiak: We need to, find* returns one record, get* returns a list.

Wendrowiec13: Right, we now have a list, my bad

PR #321, opened 2023-02-28
Display spot names on UI

adamiak:


packages/backend/src/api/controllers/TransactionSubmitController.ts
```diff
@@ -70,7 +76,7 @@ export class TransactionSubmitController {
       data: {
         type: 'Withdraw',
         starkKey: data.starkKey,
-        assetType: data.assetType,
+        assetType: AssetHash(data.assetType),
```
Wendrowiec13, line 79: The `assetType` name is really confusing. At least for me. If it's always an AssetHash we should probably change I to that

packages/backend/src/api/controllers/fetchAssetDetailsMap.ts
```diff
@@ -0,0 +1,45 @@
...
+    ...(records.assetHistory?.map((a) => a.assetHashOrId) ?? []),
+    ...(records.userTransactions?.map((t) =>
+      t.data.type === 'Withdraw'
+        ? t.data.assetType
+        : t.data.type === 'WithdrawWithTokenId' ||
+          t.data.type === 'MintWithdraw'
+        ? t.data.assetId
+        : undefined
+    ) ?? []),
+  ].filter((i) => i !== undefined) as AssetHash[]
```
Wendrowiec13, line 30: This is really hard to read. I have no idea how to make it more approachable though

packages/frontend/src/utils/assets.ts
```diff
@@ -44,6 +44,14 @@ function assetHashToInfo(
   assetDetails?: AssetDetails
 ): AssetInfo {
   if (!assetDetails) {
+    if (assetHash.startsWith('0x04')) {
+      return {
+        name: 'Mintable NFT',
+        symbol: `Mintable # ${assetHash.substring(4)}`,
+        isUnknownHash: true,
+        imageUrl: '/images/unknown-asset.svg',
+      }
```
Wendrowiec13, line 53: Why do we need that?

packages/backend/src/api/controllers/UserController.ts
```diff
@@ -81,16 +84,29 @@ export class UserController {
       this.userTransactionRepository.getCountByStarkKey(starkKey),
     ])
 
+    let assetDetailsMap = {}
```
Wendrowiec13, line 87: Could we type it as `Record<string, AssetDetails>` here?

packages/frontend/src/utils/assets.ts
adamiak: It's a temporary solution for Phase1 - we don't have any info on Mintable NFTs, so on frontend only we display a placeholder.

packages/backend/src/api/controllers/UserController.ts
adamiak: Done

packages/backend/src/api/controllers/TransactionSubmitController.ts
adamiak: I agree, but that's also how Starkware names them. It's AssetType for ERC and ETH, and AssetId for Perpetual and for Spot NFT, but also there types are different (id or hash). I agree we should unify it more on our side, but at the explorer-ethereum touchpoint we will need to use Starkware names from solidity.

PR #330, opened 2023-03-02
Implement regular withdrawals

torztomasz:
Resolves L2B-896


packages/frontend/src/view/pages/user/components/UserQuickActionsTable.tsx
```diff
@@ -50,7 +55,17 @@ export function UserQuickActionsTable(props: UserQuickActionsTableProps) {
...
               </strong>
             </p>
-            <Button className="ml-auto w-32 !px-0">Withdraw now</Button>
+            {props.user && asset.asset.details && (
+              <UserWithdrawNowButton
+                className="ml-auto w-32 !px-0"
+                assetDetails={asset.asset.details}
+                account={props.user.address}
+                starkKey={props.starkKey}
+                exchangeAddress={EthereumAddress.fake()}
```
Wendrowiec13, line 64: Is this intentional?

packages/shared/src/utils/assert.ts
```diff
@@ -0,0 +1,3 @@
+export function assertUnreachable(_: never): never {
+  throw new Error('There are more values to handle.')
+}
```
Wendrowiec13, line 3: I think we should rename this file to assertUnreachable, delete the one that's in the backend and use this one instead since it's exactly the same

packages/shared/src/utils/getAssetSelector.ts
```diff
@@ -0,0 +1,29 @@
+import { keccak256 } from '@ethersproject/solidity'
+
+import { AssetType } from '../AssetDetails'
+import { assertUnreachable } from './assert'
+
+export function getAssetSelector(
+  assetType:
+    | Extract<AssetType, 'ETH' | 'ERC20'>
+    | Extract<AssetType, 'ERC721' | 'ERC1155'>
```
Wendrowiec13, line 9: Couldn't we define it like this `Extract<AssetType, 'ERC721' | 'ERC1155' | 'ETH' | 'ERC20'>`

Wendrowiec13 (changes requested):

packages/frontend/src/view/pages/user/components/UserQuickActionsTable.tsx
torztomasz: yeah, I did not know what to put there, but it is going to be changed

packages/shared/src/utils/assert.ts
torztomasz: good catch

packages/shared/src/utils/getAssetSelector.ts
torztomasz: yup, we should

Wendrowiec13 (approved):

packages/backend/src/peripherals/database/AssetRepository.ts
```diff
@@ -70,8 +72,8 @@ export class AssetRepository extends BaseRepository {
     return row ? toAssetDetailsRecord(row) : undefined
   }
 
-  async getDetailsByAssetHashes(
-    assetHashes: AssetHash[]
+  async getDetailsByAssetHashesOrTypeHashes(
```
adamiak, line 75: Selecting by `asset_type_hash` is not necessary. `asset_details` table has ERC20 assets added with 'asset_hash' set to their type hashes. This function was fine before.

packages/backend/src/core/AssetDetailsService.ts
```diff
@@ -0,0 +1,78 @@
+import { assertUnreachable, TradingMode } from '@explorer/shared'
+import { AssetHash, Hash256 } from '@explorer/types'
+
+import { AssetRepository } from '../peripherals/database/AssetRepository'
+import { PreprocessedAssetHistoryRecord } from '../peripherals/database/PreprocessedAssetHistoryRepository'
+import { SentTransactionRecord } from '../peripherals/database/transactions/SentTransactionRepository'
+import { UserTransactionRecord } from '../peripherals/database/transactions/UserTransactionRepository'
+import { AssetDetailsMap } from './AssetDetailsMap'
+
+export class AssetDetailsService {
```
adamiak, line 10: Please add unit tests for this class

packages/backend/src/core/AssetDetailsMap.ts
```diff
@@ -0,0 +1,39 @@
+import { AssetDetails } from '@explorer/shared'
+import { AssetHash, Hash256 } from '@explorer/types'
+
+export class AssetDetailsMap {
```
adamiak, line 4: Please add unit tests for this class

packages/backend/src/core/AssetDetailsService.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+}
+
+function getSentTransactionAssetIdentifiers(
+  sentTransaction: SentTransactionRecord
+): AssetHash | Hash256 | undefined {
+  switch (sentTransaction.data.type) {
+    case 'Withdraw':
+    case 'WithdrawWithTokenId':
+      return sentTransaction.data.assetType
```
adamiak, line 53: we should search by tokenId. By ignoring tokenId here we need to fetch all possible NFTs of the same type, which is not necessary and inefficient. Let's discuss of finding a better solution,.

packages/backend/src/core/AssetDetailsService.ts
```diff
@@ -0,0 +1,78 @@
...
+function getUserTransactionAssetIdentifiers(
+  userTransaction: UserTransactionRecord
+): AssetHash | undefined {
+  switch (userTransaction.data.type) {
+    case 'Withdraw':
+      return userTransaction.data.assetType
+    case 'WithdrawWithTokenId':
+    case 'MintWithdraw':
+      return userTransaction.data.assetId
+    case 'ForcedTrade':
```
adamiak, line 71: I suggest adding a few comments here explaining why we return `undefined` for all theses three types.

packages/backend/src/core/AssetDetailsService.ts
```diff
@@ -0,0 +1,78 @@
...
+      ...(records.assetHistory?.map((a) => a.assetHashOrId) ?? []),
+      ...(records.userTransactions?.map(getUserTransactionAssetIdentifiers) ??
+        []),
+    ].filter(
+      (i): i is AssetHash | Hash256 => AssetHash.check(i) || Hash256.check(i)
+    )
+
+    const uniqueAssetHashes = [...new Set(assetHashes)]
+    const assetDetails =
+      await this.assetRepository.getDetailsByAssetHashesOrTypeHashes(
```
adamiak, line 39: See other comments, pass `tokenId`s

packages/backend/src/peripherals/database/PreprocessedAssetHistoryRepository.ts
```diff
@@ -6,7 +6,9 @@ import { Logger } from '../../tools/Logger'
 import { BaseRepository } from './shared/BaseRepository'
 import { Database } from './shared/Database'
 
-export interface PreprocessedAssetHistoryRecord<T extends AssetHash | AssetId> {
+export interface PreprocessedAssetHistoryRecord<
+  T extends AssetHash | AssetId = AssetHash | AssetId
```
adamiak, line 10: I don't understand what is happening here :) AssetId in perpetuals (e.g. "ETH-9") is different then AssetHash on spot (e.g. "0xe0ab3eef.."). Losing this distinction here can prevent TypeScript from properly ensuring types in other places. But I'm not sure if this is the case here.

packages/backend/src/peripherals/database/AssetRepository.ts
```diff
@@ -70,8 +72,8 @@ export class AssetRepository extends BaseRepository {
...
+    assetHashes: (AssetHash | Hash256)[]
```
adamiak, line 76: Asset type is not `Hash256`. It's also `AssetHash`. I know it's confusing, but they share the type.

packages/backend/src/core/AssetDetailsService.ts
torztomasz: already done

adamiak (changes requested):
Please resolve comments

packages/backend/src/core/AssetDetailsService.ts
adamiak, line 1: Do we still need `getSentTransactionAssetIdentifiers()`?

torztomasz: nope, I will delete it

adamiak (approved):

PR #331, opened 2023-03-02
Temporarily show forced action buttons only on spot

adamiak:
This is due to perpetual buttons not being properly linked.

### DIFF
--- a/packages/frontend/src/view/pages/user/UserPage.tsx
+++ b/packages/frontend/src/view/pages/user/UserPage.tsx
@@ -87,7 +87,7 @@ function UserPage(props: UserPageProps) {
             type={props.type}
             starkKey={props.starkKey}
             assets={props.assets}
-            isMine={props.user !== undefined}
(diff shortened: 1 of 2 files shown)


packages/frontend/src/view/pages/user/UserPage.tsx
```diff
@@ -87,7 +87,7 @@ function UserPage(props: UserPageProps) {
             type={props.type}
             starkKey={props.starkKey}
             assets={props.assets}
-            isMine={props.user !== undefined}
+            isMine={isMine && props.type === 'SPOT'}
```
torztomasz, line 90: I would suggest to use props.type conditional rendering inside the component not here

torztomasz (changes requested):

packages/frontend/src/view/pages/user/UserPage.tsx
adamiak: Fixed

torztomasz (approved):

PR #334, opened 2023-03-06
Implement forced actions controller

torztomasz:
Resolves L2B-950


packages/backend/src/api/controllers/ForcedActionsController.ts
```diff
@@ -0,0 +1,147 @@
...
+  ) {}
+
+  async getSpotForcedWithdrawalPage(
+    givenUser: Partial<UserDetails>,
+    vaultId: bigint
+  ): Promise<ControllerResult> {
+    const user = await this.userService.getUserDetails(givenUser)
+
+    if (!user) {
+      return { type: 'not found', content: 'User must be logged in' }
```
Wendrowiec13, line 29: Maybe we should handle those cases with some kind of error page like the 404 one? Definitely not a discussion for this PR but I think it would be a good UX addition

packages/backend/src/api/controllers/ForcedActionsController.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    const assets =
+      await this.preprocessedAssetHistoryRepository.getCurrentByPositionOrVaultId(
+        vaultId
+      )
+
+    const asset = assets[0]
```
Wendrowiec13, line 37: I think we need @adamiak ‘s opinion on this but from what he told me gammaX doesn’t follow the rule one vault = one asset. Although I may have gotten something wrong.

packages/backend/src/api/controllers/ForcedActionsController.ts
```diff
@@ -0,0 +1,147 @@
...
+    if (!asset) {
+      return {
+        type: 'not found',
+        content:
+          'Position is empty or does not contain asset with given assetId',
+      }
+    }
+
+    if (asset.starkKey != user.starkKey) {
+      return { type: 'not found', content: 'Vault does not belong to user' }
```
Wendrowiec13, line 92: Change `Vault` To `Position`

packages/backend/src/api/controllers/ForcedActionsController.ts
```diff
@@ -0,0 +1,147 @@
...
+        positionId
+      )
+
+    const asset = assets.find((asset) => asset.assetHashOrId === assetId)
+    if (!asset) {
+      return { type: 'not found', content: 'Position is empty' }
+    }
+
+    if (asset.starkKey != user.starkKey) {
+      return { type: 'not found', content: 'Vault does not belong to user' }
```
Wendrowiec13, line 131: Change `Vault` To `Position`

packages/frontend/src/view/pages/user/components/UserAssetTable.tsx
```diff
@@ -26,14 +26,21 @@ export interface UserAssetEntry {
...
+
   return (
     <Table
       fullBackground
       columns={[
         { header: <span className="pl-10">Name</span> },
         { header: 'Balance' },
         { header: props.type === 'PERPETUAL' ? 'Position' : 'Vault' },
-        { header: props.isMine ? 'Action' : '' },
+        ...(props.isMine ? [{ header: 'Action' }] : []),
```
Wendrowiec13, line 43: Why do you think that’s better?

packages/frontend/src/view/pages/user/components/UserAssetTable.tsx
```diff
@@ -55,15 +62,10 @@ export function UserAssetsTable(props: UserAssetsTableProps) {
                 <a href={`/proof/${entry.vaultOrPositionId}`}>(proof)</a>
               )}
             </span>,
-            props.isMine && props.type === 'SPOT' ? (
-              <LinkButton
-                className="w-full"
-                href={`/forced/new/spot/${entry.vaultOrPositionId}`}
-              >
+            props.isMine && (
+              <LinkButton className="w-full" href={forcedActionLink(entry)}>
```
Wendrowiec13, line 66: I am not sure whether determining whether its spot or perpetual by assetId is the way to go. Although if that’s the case maybe we could get rid of the type prop?

Wendrowiec13 (changes requested):

packages/backend/src/api/controllers/ForcedActionsController.ts
adamiak: In case of Spot there's only one asset in the vault, so it's ok here.

packages/backend/src/api/controllers/ForcedActionsController.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    const assetDetails = await this.assetRepository.findDetailsByAssetHash(
+      asset.assetHashOrId as AssetHash
+    )
+
+    const content = renderNewSpotForcedWithdrawPage({
+      user,
+      starkExAddress: EthereumAddress.ZERO,
```
adamiak, line 52: I assume this is the address of the StarkEx contract. It's in the config, although it's incorrectly named `contracts.perpetual` even for spot. Please pass the correct address here and  rename the config field (or create a story to do it later). I suggest renaming `contracts.perpetual` to `contracts.starkex` for both spot and perpetual.

packages/backend/src/api/controllers/ForcedActionsController.ts
```diff
@@ -0,0 +1,147 @@
...
+      }
+    }
+
+    if (asset.starkKey != user.starkKey) {
+      return { type: 'not found', content: 'Vault does not belong to user' }
+    }
+
+    const content = renderNewSpotForcedWithdrawPage({
+      user,
+      starkExAddress: EthereumAddress.ZERO,
```
adamiak, line 97: Pass the proper address

packages/backend/src/api/controllers/ForcedActionsController.ts
```diff
@@ -0,0 +1,147 @@
...
+      positionOrVaultId: positionId,
+      starkKey: asset.starkKey,
+      asset: {
+        hashOrId: asset.assetHashOrId,
+        balance: asset.balance,
+        priceUSDCents: 0n,
```
adamiak, line 103: Is this correct? It's a collateral, so maybe this field is not used. But I prefer to ask :)

packages/backend/src/api/controllers/ForcedActionsController.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    const content = renderNewPerpetualForcedTradePage({
+      user,
+      starkExAddress: EthereumAddress.ZERO,
+      positionOrVaultId: positionId,
+      starkKey: asset.starkKey,
+      asset: {
+        hashOrId: asset.assetHashOrId,
+        balance: asset.balance,
+        priceUSDCents: 0n,
```
adamiak, line 142: I think price needs to be properly passed for the UI to work.

adamiak (changes requested):

packages/backend/src/api/controllers/ForcedActionsController.ts
torztomasz: It is not I suppose. To be honest I do not know what should go in this place but I saw how you did the spot page and it was like that. I thought that maybe it is only temporary so decided to go with it 😅

packages/backend/src/api/controllers/ForcedActionsController.ts
torztomasz: I kinda feel that contracts.starkex is not the thing to go for because the full path would go like this `config.starkex.contracts.starkex`. Maybe we can have a call to talk about this later?

packages/frontend/src/view/pages/user/components/UserAssetTable.tsx
torztomasz: The previous approach would create column with empty header and cell, this approach will not create any column

packages/frontend/src/view/pages/user/components/UserAssetTable.tsx
torztomasz: You are right, we should use tradingMode for determining the action link

packages/frontend/src/view/pages/user/components/UserAssetTable.tsx
Wendrowiec13: I'm aware of that. Is that the desired behaviour?

adamiak (approved):

PR #344, opened 2023-03-14
Add wallet conect

vairamzen:


torztomasz:
Hey!

Thanks for contributing to L2Beat! We really appreciate the time and effort you put into it.

However, we did notice a couple of things that we'd like to address. The code quality could use some improvement, and there were some unnecessary dependencies added to the project. Additionally, while we think Wallet Connect is a neat feature, it's not something we need right now as it's not aligned with our current project goals.

vairamzen:
Hey,
                Thanks for the feedback. We need this wallet connect feature. Can you please share the changes and point out the issues in the PR. We will try to rectify those and open a new PR.
Thanks.

PR #358, opened 2023-03-22
Implement withdrawable assets

Wendrowiec13:
Resolves L2B-1040


packages/backend/src/peripherals/database/WithdrawableAssetRepository.ts
```diff
@@ -63,6 +66,19 @@ export class WithdrawableAssetRepository extends BaseRepository {
     return results[0]!.id
   }
 
+  async getWithdrawableAssetsByStarkKey(
+    starkKey: StarkKey
+  ): Promise<{ assetHash: AssetHash; balanceDelta: bigint }[]> {
+    const knex = await this.knex()
+    const results: { assetHash: AssetHash; balanceDelta: bigint }[] =
+      await knex('withdrawable_assets')
+        .select('asset_hash as assetHash')
```
torztomasz, line 75: I don't know if casting is necessary here.

packages/backend/src/peripherals/database/WithdrawableAssetRepository.ts
```diff
@@ -63,6 +66,19 @@ export class WithdrawableAssetRepository extends BaseRepository {
...
+        .where('stark_key', starkKey.toString())
+        .groupBy('asset_hash')
+        .sum('balance_delta as balanceDelta')
```
torztomasz, line 78: here also

packages/backend/src/peripherals/database/WithdrawableAssetRepository.ts
```diff
@@ -63,6 +66,19 @@ export class WithdrawableAssetRepository extends BaseRepository {
     return results[0]!.id
   }
 
+  async getWithdrawableAssetsByStarkKey(
```
torztomasz, line 69: I think it should be called getByStarkKey, it is already in WithdrawableAssetsRepository

packages/backend/src/peripherals/database/WithdrawableAssetRepository.ts
```diff
@@ -63,6 +66,19 @@ export class WithdrawableAssetRepository extends BaseRepository {
...
+    starkKey: StarkKey
+  ): Promise<{ assetHash: AssetHash; balanceDelta: bigint }[]> {
```
torztomasz, line 71: Why just returning this? Why not whole WithdrawableAssetRecord and then map it in controller?

torztomasz (changes requested):

torztomasz (approved):

packages/backend/src/peripherals/database/WithdrawableAssetRepository.test.ts
```diff
@@ -224,4 +224,89 @@ describe(WithdrawableAssetRepository.name, () => {
       expect(await repository.findById(id2)).toEqual(undefined)
     })
   })
+  describe(
+    WithdrawableAssetRepository.prototype.getAssetBalancesByStarkKey.name,
```
adamiak, line 228: Please extend the tests with scenarios where some assets sum up to zero.

packages/backend/src/api/controllers/UserController.ts
```diff
@@ -123,7 +127,10 @@ export class UserController {
       tradingMode: this.tradingMode,
       starkKey,
       ethereumAddress: registeredUser?.ethAddress ?? EthereumAddress.ZERO,
-      withdrawableAssets: [],
+      withdrawableAssets: withdrawableAssets.map((asset) => ({
+        asset: { hashOrId: asset.assetHash },
```
adamiak, line 131: Don't we need full asset details here to properly display its name and amount? I think it's necessary to reach to `assetDetailsMap` first to get it.

packages/backend/src/api/controllers/UserController.ts
```diff
@@ -83,6 +86,7 @@ export class UserController {
         limit: 10,
       }),
       this.userTransactionRepository.getCountByStarkKey(starkKey),
+      this.withdrawableAssetRepository.getAssetBalancesByStarkKey(starkKey),
     ])
 
     const assetDetailsMap = await getAssetHashToAssetDetailsMap(
```
adamiak, line 92: don't we need to pass withdrawable assets here to fetch their details?

adamiak (changes requested):

packages/backend/src/peripherals/database/WithdrawableAssetRepository.ts
```diff
@@ -63,6 +66,19 @@ export class WithdrawableAssetRepository extends BaseRepository {
...
+    const knex = await this.knex()
+    const results: { assetHash: AssetHash; balanceDelta: bigint }[] =
+      await knex('withdrawable_assets')
+        .select('asset_hash as assetHash')
+        .where('stark_key', starkKey.toString())
+        .groupBy('asset_hash')
+        .sum('balance_delta as balanceDelta')
+    return results.filter((result) => result.balanceDelta > 0)
```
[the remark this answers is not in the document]
Wendrowiec13: Why is filtering with SQL better?

packages/backend/src/peripherals/database/WithdrawableAssetRepository.ts
```diff
@@ -63,6 +66,19 @@ export class WithdrawableAssetRepository extends BaseRepository {
     return results[0]!.id
   }
 
+  async getAssetBalancesByStarkKey(
```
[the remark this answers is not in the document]
Wendrowiec13: I would actually like to rename the repository because WithdrawableAssetRespoitory doesn't contain withdrawable assets and then I would change the name to the initial one which was `getWithdrawableAssetsByStarkKey`

packages/backend/src/peripherals/database/WithdrawableAssetRepository.ts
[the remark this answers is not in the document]
adamiak: Because otherwise the database returns values for all users from the beginning of time, even though sums of 99.999% of them are 0. So you get big response from DB that you later filter in JS. It would be more efficient to just get the records you're interested in.

packages/backend/src/peripherals/database/WithdrawableAssetRepository.ts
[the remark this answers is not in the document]
adamiak: What would you rename the repo to?

adamiak (approved):
Approved, but please rename as requested in the comment.

PR #385, opened 2023-04-16
Extract withdrawables collection process out of user transaction migrator

adamiak:
Resolves L2B-946

To make things simple and safe, a new `WithdrawableAssetMigrator` was added to take care of the process that was incorrectly put into `UserTransactionMigrator`.


torztomasz (approved):
LGTM, although the only thing I dislike is the UserTransactionCollector options. What would you say about separating logic from UserTransactionCollector to ForcedTransactionCollector and WithdrawalCollector?

PR #386, opened 2023-04-16
Add .devcontainer

adamiak:
The `.devcontainer` creates both the Linux system for development, but also sets up a PostgreSQL database.


.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+    # Use "forwardPorts" in **devcontainer.json** to forward an app port locally. 
+    # (Adding the "ports" property to this file will not forward from a Codespace.)
+
+  db:
+    image: postgres:14
+    restart: unless-stopped
+    volumes:
+      - postgres-data:/var/lib/postgresql/data
+    environment:
+      POSTGRES_PASSWORD: postgres
```
torztomasz, line 35: Should'nt it be "password"? 
`docker run -d --name=state_explorer_postgres -p 5432:5432 -e POSTGRES_PASSWORD=password postgres` That's the command for creating backend database.

torztomasz (approved):

.devcontainer/docker-compose.yml
adamiak: It's correct - that's the standard setup from VSCode if you build the container with postgres from scratch.

PR #394, opened 2023-04-20
enable favicon

torztomasz:


adamiak:
It does seem like the favico works fine and this is not needed.

PR #404, opened 2023-05-25
implement TransactionDownloader

torztomasz:
Resolves L2B-1539

### DIFF
--- a/packages/backend/src/peripherals/database/L2TransactionRepository.ts
+++ b/packages/backend/src/peripherals/database/L2TransactionRepository.ts
@@ -19,7 +19,7 @@ interface Record<
   stateUpdateId: number
   blockNumber: number
   parentId: number | undefined
-  status: 'alternative' | 'replaced' | undefined
+  state: 'alternative' | 'replaced' | undefined
   starkKeyA: StarkKey | undefined
   starkKeyB: StarkKey | undefined
   type: T
@@ -54,7 +54,7 @@ export class L2TransactionRepository extends BaseRepository {
 
     if (count === 1) {
       await knex('l2_transactions')
-        .update({ status: 'replaced' })
+        .update({ state: 'replaced' })
         .where({ transaction_id: record.transactionId })
     }
 
@@ -90,7 +90,7 @@ export class L2TransactionRepository extends BaseRepository {
         state_update_id: record.stateUpdateId,
         block_number: record.blockNumber,
         parent_id: record.parentId,
-        status: record.isAlternative ? 'alternative' : null,
+        state: record.isAlternative ? 'alternative' : null,
         stark_key_a: starkKeyA?.toString(),
         stark_key_b: starkKeyB?.toString(),
         type: record.data.type,
@@ -119,7 +119,7 @@ export class L2TransactionRepository extends BaseRepository {
         state_update_id: record.stateUpdateId,
         block_number: record.blockNumber,
         type: record.data.type,
-        status: record.isAlternative ? 'alternative' : null,
+        state: record.isAlternative ? 'alternative' : null,
         data,
       })
       .returning('id')
@@ -189,7 +189,7 @@ function toRecord(row: L2TransactionRow): Record {
     stateUpdateId: row.state_update_id,
     blockNumber: row.block_number,
     parentId: row.parent_id ? row.parent_id : undefined,
-    status: row.status ? row.status : undefined,
+    state: row.state ? row.state : undefined,
     starkKeyA: row.stark_key_a ? StarkKey(row.stark_key_a) : undefined,
     starkKeyB: row.stark_key_b ? StarkKey(row.stark_key_b) : undefined,
     type: row.type as L2TransactionData['type'],
--- a/packages/backend/src/peripherals/database/L2TransactionRepository.test.ts
+++ b/packages/backend/src/peripherals/database/L2TransactionRepository.test.ts
@@ -39,7 +39,7 @@ describe(L2TransactionRepository.name, () => {
         starkKeyB: undefined,
         type: record.data.type,
         data: record.data,
-        status: undefined,
+        state: undefined,
         parentId: undefined,
       })
     })
@@ -79,7 +79,7 @@ describe(L2TransactionRepository.name, () => {
         starkKeyB: undefined,
         type: record.data.type,
         data: record.data,
-        status: undefined,
+        state: undefined,
         parentId: undefined,
       })
 
@@ -96,7 +96,7 @@ describe(L2TransactionRepository.name, () => {
         starkKeyB: undefined,
         type: record.data.type,
         data: record.data,
-        status: 'replaced',
+        state: 'replaced',
         parentId: undefined,
       })
 
@@ -108,7 +108,7 @@ describe(L2TransactionRepository.name, () => {
         starkKeyB: undefined,
         type: alternativeRecord.data.type,
         data: alternativeRecord.data,
-        status: 'alternative',
+        state: 'alternative',
         parentId: undefined,
       })
     })
@@ -153,7 +153,7 @@ describe(L2TransactionRepository.name, () => {
         stateUpdateId: record.stateUpdateId,
         transactionId: record.transactionId,
         blockNumber: record.blockNumber,
-        status: undefined,
+        state: undefined,
         parentId: undefined,
         starkKeyA: undefined,
         starkKeyB: undefined,
@@ -164,7 +164,7 @@ describe(L2TransactionRepository.name, () => {
         stateUpdateId: record.stateUpdateId,
         transactionId: record.transactionId,
         blockNumber: record.blockNumber,
-        status: undefined,
+        state: undefined,
         parentId: id,
         starkKeyA: StarkKey.fake('1'),
         starkKeyB: undefined,
@@ -175,7 +175,7 @@ describe(L2TransactionRepository.name, () => {
         stateUpdateId: record.stateUpdateId,
         transactionId: record.transactionId,
         blockNumber: record.blockNumber,
-        status: undefined,
+        state: undefined,
         parentId: id,
         starkKeyA: StarkKey.fake('2'),
         starkKeyB: undefined,
@@ -263,7 +263,7 @@ describe(L2TransactionRepository.name, () => {
         starkKeyB: undefined,
         type: record.data.type,
         data: record.data,
-        status: 'replaced',
+        state: 'replaced',
         parentId: undefined,
       })
       expect(firstTransactionOfMultiAfterAlternative).toEqual({
@@ -274,7 +274,7 @@ describe(L2TransactionRepository.name, () => {
         starkKeyB: undefined,
         type: record.data.transactions[0]!.type,
         data: record.data.transactions[0]!,
-        status: 'replaced',
+        state: 'replaced',
         parentId: id,
       })
       expect(secondTransactionOfMultiAfterAlternative).toEqual({
@@ -285,7 +285,7 @@ describe(L2TransactionRepository.name, () => {
         starkKeyB: undefined,
         type: record.data.transactions[1]!.type,
         data: record.data.transactions[1]!,
-        status: 'replaced',
+        state: 'replaced',
         parentId: id,
       })
 
@@ -301,7 +301,7 @@ describe(L2TransactionRepository.name, () => {
         starkKeyB: undefined,
         type: alternativeRecord.data.type,
         data: alternativeRecord.data,
-        status: 'alternative',
+        state: 'alternative',
         parentId: undefined,
       })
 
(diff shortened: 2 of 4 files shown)


packages/backend/src/core/PerpetualValidiumSyncService.ts
```diff
@@ -19,6 +20,7 @@ export class PerpetualValidiumSyncService implements IDataSyncService {
     private readonly perpetualCairoOutputCollector: PerpetualCairoOutputCollector,
     private readonly perpetualValidiumUpdater: PerpetualValidiumUpdater,
     private readonly withdrawalAllowedCollector: WithdrawalAllowedCollector,
+    private readonly transactionDownloader: TransactionDownloader | undefined,
```
adamiak, line 23: I suggest using "?" instead of "| undefined"

packages/backend/src/peripherals/starkware/TransactionDownloader.ts
```diff
@@ -0,0 +1,54 @@
...
+    private readonly feederGatewayClient: FeederGatewayClient,
+    private readonly stateUpdateRepository: StateUpdateRepository,
+    private readonly transactionRepository: TransactionRepository,
+    private readonly logger: Logger
+  ) {}
+
+  async sync(syncToBatchId: number) {
+    const latestSyncedTransactionStateUpdateId =
+      await this.transactionRepository.findLatestStateUpdateId()
+    const startSyncFromBatchId = latestSyncedTransactionStateUpdateId
```
adamiak, line 17: Use "??"

packages/backend/src/peripherals/starkware/TransactionDownloader.ts
```diff
@@ -0,0 +1,54 @@
...
+        stateUpdateId,
+      })
+
+      const data = await this.feederGatewayClient.getPerpetualBatchInfo(batchId)
+
+      const stateUpdate = await this.stateUpdateRepository.findById(
+        stateUpdateId
+      )
+      if (!stateUpdate) {
+        throw new Error(`State update ${stateUpdateId} not found`)
```
adamiak, line 37: This should never happen, see comment above on the for loop.

packages/backend/src/peripherals/starkware/TransactionDownloader.ts
```diff
@@ -0,0 +1,54 @@
...
+      }
+
+      for (const transactionInfo of data.transactionsInfo) {
+        await this.transactionRepository.add({
```
adamiak, line 41: Will this add all alternatives? We should add alternatives to the table as well.

adamiak (changes requested):

packages/backend/src/core/PerpetualValidiumSyncService.ts
torztomasz: Well, I would say that `| undefined` is a better pattern here. This way we enforce on someone to specify it, making sure that someone will not forget to pass it.

packages/backend/src/peripherals/database/L2TransactionRepository.ts
```diff
@@ -44,7 +44,7 @@ export class L2TransactionRepository extends BaseRepository {
     const { starkKeyA, starkKeyB, data } = encodeL2TransactionData(record.data)
 
     const count = await this.countByTransactionId(record.transactionId)
```
adamiak, line 51: I'm worried that this is too hard on database, especially that we will be getting high load of transactions. Adding needs to be fast.

So either: pass altIndex to the function, or revisit idea of primary key.

adamiak (changes requested):
See comment to L2TransactionRepository.

packages/backend/src/peripherals/database/L2TransactionRepository.ts
torztomasz: We can not pass altIndex to function in case of transaction API.

adamiak: in this case I'm leaning towards using primary key. Calling count() and then insert() in sequence is inefficient, and unsafe in case of asynchronous processes modifying this table (e.g. we don't have any guarantee that between calculating count and adding row, the count doesn't change). I know it's all edge cases and fast enough, but given two options:

1. using primary key = simple and easy code, no race conditions
2. using alt_id = more complex, inefficient code with race conditions

why would we choose 2? :)

adamiak: now I've realised that we can't escape looking into the table before adding a new record, because we need to mark the new record as a replacement and the previous one as replaced. So solution 1. above isn't simpler.

adamiak:
> @adamiak I am wondering if the name should be TransactionDownloader. I did not like FeederGatewayCollector as it refers mainly to the logs (that are collected, we are rather downloading stuff).

Personally FeederGatewayCollector  sounds more meaningful to me. I know it's a collector (so it follows the collector pattern) and I know it's downloading from feeder gateway. With `TransactionDownloader` I don't know which transactions, where from, and how is this "Downloader" implemented.  I'll leave the comment in the code.

But I agree that your name is "cleaner". I just prioritize having the name being very meaningful and convey as much technical details as possible.

packages/backend/src/peripherals/starkware/L2TransactionDownloader.ts
```diff
@@ -0,0 +1,56 @@
+import { Logger } from '../../tools/Logger'
+import { L2TransactionRepository } from '../database/L2TransactionRepository'
+import { StateUpdateRecord } from '../database/StateUpdateRepository'
+import { FeederGatewayClient } from './FeederGatewayClient'
+
+export class L2TransactionDownloader {
```
adamiak, line 6: I don't see how it's any different than all other collectors we have.

I suggest renaming to FeederGatewayCollector, renaming "sync" to "collect" and moving to "/collectors" folder.

packages/backend/src/peripherals/starkware/L2TransactionDownloader.ts
```diff
@@ -0,0 +1,56 @@
...
+    stateUpdates: Pick<StateUpdateRecord, 'id' | 'batchId' | 'blockNumber'>[]
+  ) {
+    const latestSyncedTransactionStateUpdateId =
+      (await this.transactionRepository.findLatestStateUpdateId()) ?? 0
+
+    for (const stateUpdate of stateUpdates) {
+      if (stateUpdate.id <= latestSyncedTransactionStateUpdateId) {
+        continue
+      }
+      this.logger.info(`Syncing transactions...`, {
```
adamiak, line 23: Please be more explicit, e.g. "Downloading transactions from Feeder Gateway..."

packages/backend/src/peripherals/starkware/L2TransactionDownloader.ts
```diff
@@ -0,0 +1,56 @@
...
+        await this.transactionRepository.add({
+          stateUpdateId: stateUpdate.id,
+          blockNumber: stateUpdate.blockNumber,
+          transactionId: transactionInfo.originalTransactionId,
+          data: transactionInfo.originalTransaction,
+        })
+        if (!transactionInfo.alternativeTransactions) {
+          continue
+        }
+        for (const alternativeTransaction of transactionInfo.alternativeTransactions) {
```
adamiak, line 41: Rather than two loops doing the same, it should be nicer to just create a new array:

```
const transactions = [data.transactionInfo, ...data.transactionInfo.alternativeTransactions || []]
````

(or something similar), and then have just one loop through it.

File: packages/backend/src/peripherals/starkware/L2TransactionDownloader.ts [hunk 9]
```diff
@@ -0,0 +1,56 @@
...
+
+export class L2TransactionDownloader {
+  constructor(
+    private readonly feederGatewayClient: FeederGatewayClient,
+    private readonly transactionRepository: L2TransactionRepository,
+    private readonly logger: Logger
+  ) {}
+
+  async sync(
+    stateUpdates: Pick<StateUpdateRecord, 'id' | 'batchId' | 'blockNumber'>[]
```
adamiak, line 14: I don't understand why we need a list here. I think we just need the latest state update id to which want to collect l2 transacations...

packages/backend/src/peripherals/starkware/L2TransactionDownloader.ts
```diff
@@ -0,0 +1,56 @@
...
+  ) {
+    const latestSyncedTransactionStateUpdateId =
+      (await this.transactionRepository.findLatestStateUpdateId()) ?? 0
+
+    for (const stateUpdate of stateUpdates) {
```
adamiak, line 19: We should start syncing from `latestSyncedTransactionStateUpdateId` which can be 0. We should not just loop stateUpdates. I don't think we need a list of stateUpdates, just the latest id....

packages/backend/src/core/PerpetualValidiumSyncService.ts
```diff
@@ -42,23 +50,34 @@ export class PerpetualValidiumSyncService implements IDataSyncService {
       userRegistrations: userRegistrations.length,
     })
 
-    for (const transition of stateTransitions) {
+    const stateUpdates = await this.processStateUpdates(stateTransitions)
+    await this.l2TransactionDownloader?.sync(stateUpdates)
```
adamiak, line 54: I don't think we need to get a list of stateUpdates processed here. We just want the newest stateUpdateId and then the `l2TransactionDownloader` should sync up to that.

packages/backend/src/peripherals/starkware/L2TransactionDownloader.ts
```diff
@@ -0,0 +1,56 @@
...
+      if (stateUpdate.id <= latestSyncedTransactionStateUpdateId) {
+        continue
+      }
+      this.logger.info(`Syncing transactions...`, {
+        stateUpdateId: stateUpdate.id,
+      })
+
+      const data = await this.feederGatewayClient.getPerpetualBatchInfo(
```
adamiak, line 27: I'm pretty sure sometimes feederGatway will not have the data for the stateUpdateId we ask for. We need to handle this case gracefully and stop the collection process.

packages/backend/src/peripherals/starkware/L2TransactionDownloader.ts
```diff
@@ -0,0 +1,56 @@
...
+            transactionId: transactionInfo.originalTransactionId,
+            data: alternativeTransaction,
+          })
+        }
+      }
+    }
+  }
+
+  async discardAfter(blockNumber: number) {
+    await this.transactionRepository.deleteAfterBlock(blockNumber)
```
adamiak, line 54: I think we also need to delete all that have `blockNumber` undefined (from live API), otherwise there will be a gap between feeder data and newer live transactions.

adamiak (changes requested):

packages/backend/src/core/PerpetualValidiumSyncService.ts
torztomasz: If we can get it from processStateUpdates then why would query database later for batchId etc.?

packages/backend/src/peripherals/starkware/L2TransactionDownloader.ts
torztomasz: This will not work as we would need to do something like this: 

`const transactions = [data.transactionInfo.originalTransaction, ...data.transactionInfo.alternativeTransactions ?? []]`

And in the end, we need transactionId which is included in transactionInfo that we would not have access to in a for loop of the above array

packages/backend/src/core/collectors/PerpetualRollupStateTransitionCollector.ts
```diff
@@ -8,32 +8,76 @@ import {
...
   async collect(
-    blockRange: BlockRange
-  ): Promise<Omit<StateTransitionRecord, 'id'>[]> {
+    blockRange: BlockRange,
+    skipAddingToDb = false
+  ): Promise<PerpetualRollupStateTransition[]> {
     const logs = await this.ethereumClient.getLogsInRange(blockRange, {
       address: this.perpetualAddress.toString(),
-      topics: [LogStateTransitionFact.topic],
+      topics: [LogStateTransitionFact.topic, LogUpdateState.topic],
```
adamiak, line 30: Do we have this event in rollup mode?

adamiak: Please double-check by running on any dydx setup.

packages/backend/src/peripherals/database/L2TransactionRepository.ts
```diff
@@ -0,0 +1,200 @@
...
+    record: {
+      transactionId: number
+      stateUpdateId: number
+      blockNumber: number
+      isAlternative: boolean
+      data: MultiL2TransactionData
+    },
+    knex: Knex
+  ) {
+    console.log('addMultiTransaction', record)
```
adamiak, line 114: console.log :)

adamiak (approved):

PR #413, opened 2023-06-19
Skip dataAvailabilityMode in program output

adamiak:
This is a fix for Apex Goerli testnet, which switches to a different program output (one more field) at block 8056029.


packages/encoding/src/decoding/decodePerpetualCairoOutput.ts
```diff
@@ -8,9 +8,16 @@ import { readForcedActions } from './readForcedActions'
...
 
-export function decodePerpetualCairoOutput(data: string): PerpetualCairoOutput {
+export function decodePerpetualCairoOutput(
+  data: string,
+  skipDataAvailabilityModeField?: boolean
+): PerpetualCairoOutput {
   const reader = new ByteReader(data)
 
+  if (skipDataAvailabilityModeField) {
+    const _dataAvailabilityMode = reader.readBigInt(32)
```
torztomasz, line 18:
```suggestion
    reader.readBigInt(32)
```

packages/backend/src/core/collectors/PerpetualCairoOutputCollector.ts
```diff
@@ -13,16 +14,31 @@ const ABI = new utils.Interface([
...
     }
     const decoded = ABI.decodeFunctionData('updateState', tx.data)
     const programOutputValues = decoded.programOutput as BigNumber[]
     const hexData = toHexData(programOutputValues)
-    return decodePerpetualCairoOutput(hexData)
+
+    // This is a fix for Apex Goerli testnet, which switches
+    // to a different program output (one more field) at block 8056029.
+    // See: https://github.com/starkware-libs/stark-perpetual/blob/eaa284683deec190407fece98b96546d10f6ad67/src/services/perpetual/cairo/output/program_output.cairo#L36
+    const skipDataAvailabilityModeField =
```
torztomasz, line 37: I would extract it to separate method called `shouldSkipDataAvailabilityModeField` so it does not make mess in collect function. The comment could be moved there also.

torztomasz (changes requested):

packages/backend/src/core/collectors/PerpetualCairoOutputCollector.ts
adamiak: Done

packages/backend/src/core/collectors/shouldSkipDataAvailabilityModeField.ts
```diff
@@ -0,0 +1,16 @@
+import { InstanceName } from '@explorer/shared'
+
+export function shouldSkipDataAvailabilityModeField(
```
torztomasz, line 3: In my opinion this function should be inside PerpetualCairoOutputCollector.ts

torztomasz (approved):

PR #421, opened 2023-07-11
Implement Escape Hatch functionality

adamiak:
Resolves L2B-1621, L2B-1620, L2B-1619, L2B-1618, L2B-1617, L2B-1616


packages/backend/src/peripherals/database/KeyValueStore.ts
```diff
@@ -37,6 +45,15 @@ export class KeyValueStore extends BaseRepository {
     /* eslint-enable @typescript-eslint/unbound-method */
   }
 
+  async findByKeyWithDefault<K extends KeyValueRecord['key']>(
```
torztomasz, line 48: well I would extend the findByKey method with fallback arg instead creating new function for it

packages/backend/src/peripherals/database/StateUpdateRepository.ts
```diff
@@ -291,3 +302,41 @@ function toStateUpdatePriceRecord(
     price: record.price,
   }
 }
+
+function encodeState(state: State): ToJSON<State> {
```
torztomasz, line 306: First thought was that this fn should be in encoding package, but clearly it should not. I would change the name encode/decode to serialize/deserialize or smth

packages/frontend/src/scripts/forced-actions/escape/finalizeEscapeForm.ts
```diff
@@ -0,0 +1,43 @@
...
+  // // const hash = await Wallet.sendVerifyEscapeTransaction(
+  await Wallet.sendFinalizeEscapeTransaction(
+    user.address,
+    props.ownerStarkKey,
+    props.positionOrVaultId,
+    props.amount,
+    props.exchangeAddress
+  )
+
+  // // TODO: should we save via the API to our DB?
```
torztomasz, line 40: Hmm, thats something to decide

packages/frontend/src/view/components/page/Navbar.tsx
```diff
@@ -16,49 +16,68 @@ export function Navbar({ searchBar = true, context }: NavbarProps) {
...
+          <a href="/freeze" className="underline">
+            Read more
+          </a>
+        </div>
+      )}
+      {context.freezeStatus === 'frozen' && (
+        <div className="flex items-center justify-center gap-2 bg-red-500 px-6 py-0.5 text-white">
+          <span>This exchange is FROZEN and no longer operates normally. </span>
+        </div>
+      )}
```
torztomasz, line 80: I would extract it to component called `FreezeBanner` pass there `context.freezeStatus` as prop and render these banners conditionally inside.

packages/frontend/src/view/pages/user/components/UserAssetTable.tsx
```diff
@@ -34,6 +35,15 @@ export function UserAssetsTable(props: UserAssetsTableProps) {
         }/${entry.asset.hashOrId.toString()}`
       : `/forced/new/${entry.vaultOrPositionId}`
 
+  const escapeHatchElem = (entry: UserAssetEntry) =>
+    entry.action === 'WITHDRAW' ? (
+      <LinkButton href={`/escape/${entry.vaultOrPositionId}`}>
+        ESCAPE
+      </LinkButton>
+    ) : (
+      <span className="text-zinc-500">use collateral escape</span>
```
torztomasz, line 44: That would require improvement

packages/frontend/tsconfig.build.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "./tsconfig.json",
-  "include": ["src/view"],
+  "include": ["src"],
```
torztomasz, line 3: Why is this change required?

torztomasz (changes requested):

torztomasz:
I think we should have a way to get notified about instance being frozen. We can use rollbar.critical for this one. We would have to extend Logger functionality but it should not be hard.

packages/backend/src/api/controllers/EscapeHatchController.ts
```diff
@@ -0,0 +1,129 @@
...
+    const oldestNotIncludedForcedAction =
+      await this.freezeCheckService.findOldestNotIncludedForcedAction()
+    if (!oldestNotIncludedForcedAction) {
+      return {
+        type: 'not found',
+        message: 'Freeze is no longer possible',
+      }
+    }
+
+    if (oldestNotIncludedForcedAction.data.type !== 'ForcedWithdrawal') {
```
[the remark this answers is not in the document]
adamiak: You're right. Freezing by Forced Trade is part of a separate issue: L2B-1957

packages/backend/src/api/controllers/TransactionController.ts
```diff
@@ -183,6 +183,9 @@ export class TransactionController {
         })
         return { type: 'success', content }
       }
+      case 'EscapeVerified': {
+        return { type: 'not found', message: 'NOT IMPLEMENTED' }
```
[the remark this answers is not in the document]
adamiak: Yes, there's a separate issue for this: L2B-1931

packages/backend/src/api/controllers/UserController.ts
```diff
@@ -459,6 +455,76 @@ export class UserController {
     })
     return { type: 'success', content }
   }
+
+  async getEscapableAssets(
```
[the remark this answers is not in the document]
adamiak: done

packages/backend/src/api/routers/FrontendRouter.ts
```diff
@@ -29,6 +30,7 @@ export function createFrontendRouter(
   merkleProofController: MerkleProofController,
   searchController: SearchController,
   l2TransactionController: L2TransactionController,
+  escapeActionController: EscapeHatchController,
```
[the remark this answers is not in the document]
adamiak: fixed

packages/backend/src/core/FreezeCheckService.ts
```diff
@@ -0,0 +1,122 @@
...
+import { Logger } from '@l2beat/backend-tools'
+import { ethers } from 'ethers'
+
+import { KeyValueStore } from '../peripherals/database/KeyValueStore'
+import { UserTransactionRepository } from '../peripherals/database/transactions/UserTransactionRepository'
+import { EthereumClient } from '../peripherals/ethereum/EthereumClient'
+
+export class FreezeCheckService {
+  constructor(
+    private readonly perpatualContractAddress: EthereumAddress,
```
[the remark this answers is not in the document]
adamiak: fxied

packages/backend/src/peripherals/database/KeyValueStore.ts
adamiak: That was my initial implementation, but then it makes two last parameters of `findByKey` optional: `defaultValue` and `trx`. Then you either need to pass `undefined` if you want to use transaction but no default, or the other way around. Both look ugly. That's why I went for a separate method.

packages/backend/src/peripherals/database/StateUpdateRepository.ts
adamiak: I renamed to `stateToJson` and `stateFromJson`

packages/frontend/src/scripts/forced-actions/escape/finalizeEscapeForm.ts
adamiak: Good point, I knew about it but forgot to create a story. Here it is: L2B-1978

packages/frontend/src/view/pages/user/components/UserAssetTable.tsx
adamiak: I'll discuss this with Janek (as all the other UIs I did myself)

packages/frontend/tsconfig.build.json
adamiak: yarn build was not catching obvious syntax errors in tsx files.

adamiak:
> I think we should have a way to get notified about instance being frozen. We can use rollbar.critical for this one. We would have to extend Logger functionality but it should not be hard.

Good idea. I added a separate story for it: L2B-1979

packages/backend/src/api/controllers/EscapeHatchController.ts
```diff
@@ -0,0 +1,129 @@
...
+    if (!latestStateUpdate.perpetualState) {
+      throw new Error('No perpetual state recorded to perform escape')
+    }
+    const encodedState = encodeState(latestStateUpdate.perpetualState)
+    const serializedState: bigint[] = []
+    for (let i = 0; i < encodedState.length / 2 / 32; i++) {
+      serializedState.push(
+        BigInt('0x' + encodedState.slice(64 * i, 64 * (i + 1)))
+      )
+    }
```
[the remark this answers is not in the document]
adamiak: Moved to encoding - we only use it to serialize encoded state.

packages/frontend/src/view/components/page/FreezeBanner.tsx
```diff
@@ -0,0 +1,25 @@
...
+    )
+  }
+  if (freezeStatus === 'frozen') {
+    return (
+      <div className="flex items-center justify-center gap-2 bg-red-500 px-6 py-0.5 text-white">
+        <span>This exchange is FROZEN and no longer operates normally. </span>
+      </div>
+    )
+  }
+  return <></>
```
torztomasz, line 24: its better to return null

packages/backend/src/api/controllers/getEscapableAssets.ts
```diff
@@ -0,0 +1,79 @@
+import { EscapableAssetEntry } from '@explorer/frontend'
+import { CollateralAsset, PageContext } from '@explorer/shared'
+import { StarkKey } from '@explorer/types'
+import uniqBy from 'lodash/uniqBy'
+
+import { UserTransactionRepository } from '../../peripherals/database/transactions/UserTransactionRepository'
+import { WithdrawableAssetRepository } from '../../peripherals/database/WithdrawableAssetRepository'
+import { getCollateralAssetDetails } from './getCollateralAssetDetails'
+
+export async function getEscapableAssets(
```
torztomasz, line 10: I'm not sure if separate file is a good place, maybe we can create `WithdrawableAssetService` and put there the stark key/eth address logic behind withdrawable asset and this function?

torztomasz (changes requested):

packages/backend/src/peripherals/database/StateUpdateRepository.ts
```diff
@@ -253,7 +253,7 @@ function toStateUpdateRecord(row: StateUpdateRow): StateUpdateRecord {
     timestamp: Timestamp(row.timestamp),
     perpetualState:
       row.perpetual_state !== null
-        ? decodeState(row.perpetual_state)
+        ? stateFromJson(row.perpetual_state)
```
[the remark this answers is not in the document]
adamiak: I'm not sure if the convention works here. We're not serializing here (serializing it means turning into a sequence of something). Here we are turning into json to put into a Json column.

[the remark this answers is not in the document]
adamiak: Also, I don't see conventions to call such functions "serialize" in other *Repositories.

torztomasz (approved):

PR #422, opened 2023-07-11
Implement l2 transactions preprocessing

torztomasz:
Resolves L2B-1752

### DIFF
--- a/packages/backend/src/core/preprocessing/StateDetailsPreprocessor.ts
+++ b/packages/backend/src/core/preprocessing/StateDetailsPreprocessor.ts
@@ -1,17 +1,20 @@
 import { Knex } from 'knex'
 
+import { L2TransactionRepository } from '../../peripherals/database/L2TransactionRepository'
 import { PreprocessedAssetHistoryRepository } from '../../peripherals/database/PreprocessedAssetHistoryRepository'
 import { PreprocessedStateDetailsRepository } from '../../peripherals/database/PreprocessedStateDetailsRepository'
 import { StateUpdateRecord } from '../../peripherals/database/StateUpdateRepository'
 import { UserTransactionRepository } from '../../peripherals/database/transactions/UserTransactionRepository'
 import { Logger } from '../../tools/Logger'
+import { sumNumericValuesByKey } from '../../utils/sumNumericValuesByKey'
 
 export class StateDetailsPreprocessor {
   constructor(
-    protected readonly preprocessedStateDetailsRepository: PreprocessedStateDetailsRepository,
-    protected readonly preprocessedAssetHistoryRepository: PreprocessedAssetHistoryRepository,
-    protected readonly userTransactionRepository: UserTransactionRepository,
-    protected readonly logger: Logger
+    private readonly preprocessedStateDetailsRepository: PreprocessedStateDetailsRepository,
+    private readonly preprocessedAssetHistoryRepository: PreprocessedAssetHistoryRepository,
+    private readonly userTransactionRepository: UserTransactionRepository,
+    private readonly l2TransactionRepository: L2TransactionRepository,
+    private readonly logger: Logger
   ) {}
 
   async preprocessNextStateUpdate(
@@ -41,6 +44,48 @@ export class StateDetailsPreprocessor {
       },
       trx
     )
+  }
+
+  async catchUpL2Transactions(
+    trx: Knex.Transaction,
+    preprocessToStateUpdateId: number
+  ) {
+    const recordsToUpdate =
+      await this.preprocessedStateDetailsRepository.getAllWithoutL2TransactionStatisticsUpToStateUpdateId(
+        preprocessToStateUpdateId,
+        trx
+      )
+
+    for (const recordToUpdate of recordsToUpdate) {
+      this.logger.info(
+        `Catching up L2 transaction statistics for state update ${recordToUpdate.stateUpdateId}`
+      )
+
+      const statisticsForStateUpdate =
+        await this.l2TransactionRepository.getStatisticsByStateUpdateId(
+          recordToUpdate.stateUpdateId,
+          trx
+        )
+      const mostRecentPreprocessedStateDetails =
+        await this.preprocessedStateDetailsRepository.findMostRecentWithL2TransactionStatistics(
+          trx
+        )
+
+      await this.preprocessedStateDetailsRepository.update(
+        {
+          id: recordToUpdate.id,
+          l2TransactionsStatistics: statisticsForStateUpdate,
+          cumulativeL2TransactionsStatistics:
+            mostRecentPreprocessedStateDetails?.cumulativeL2TransactionsStatistics
+              ? sumNumericValuesByKey(
+                  mostRecentPreprocessedStateDetails.cumulativeL2TransactionsStatistics,
+                  statisticsForStateUpdate
+                )
+              : statisticsForStateUpdate,
+        },
+        trx
+      )
+    }
   }
 
   async rollbackOneStateUpdate(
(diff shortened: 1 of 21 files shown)


packages/backend/src/core/preprocessing/UserStatisticsPreprocessor.ts
```diff
@@ -117,6 +116,51 @@ export class UserStatisticsPreprocessor {
     }
   }
 
+  async catchUpL2Transactions(
+    trx: Knex.Transaction,
+    preprocessToStateUpdateId: number
+  ) {
+    const recordsToUpdate =
+      await this.preprocessedUserStatisticsRepository.getAllWithoutL2TransactionStatisticsUpToStateUpdateId(
```
adamiak, line 124: I'm afraid that this approach is incorrect. What it does is it gets state update statistics for each user, and then updates them with l2 statistics for those users. 

But what if a user buys 1 ETH and then sells 1 ETH in the same state update. There will be no entry for this user in the state update (their balance didn't change), even though he made 2 transactions. With current implementation it will miss these 2 transactions completely. 

It seems that proper implementation should go straight to l2_transactions table for given state update, fetch unique set of users (starkKeys) and then calculate statistics for each user. Those statisics should be written to existing statistics record, or a new record needs to be added with zero's in columns for state update statistics.

So either implement as suggested above, or let's have a separate table with only l2 preprocessed statistics... Let me know what you prefer.

packages/backend/src/peripherals/database/PreprocessedUserStatisticsRepository.ts
```diff
@@ -50,7 +56,47 @@ export class PreprocessedUserStatisticsRepository extends BaseRepository {
...
+    const row = await knex('preprocessed_user_statistics')
+      .whereNotNull('l2_transactions_statistics')
+      .andWhere('stark_key', starkKey.toString())
+      .orderBy('state_update_id', 'desc')
+      .first()
+
+    return row ? toPreprocessedUserStatisticsRecord(row) : undefined
+  }
+
+  async getAllWithoutL2TransactionStatisticsUpToStateUpdateId(
```
adamiak, line 76: Check my other comments, but this function probably is not necessary at all, because we can't depend on state update statistics for L2 transaction preprocessing, because there are scenarios when user had no state update but he has transactions (i.e. -1 and then +1)

packages/backend/src/test/fakes.ts
```diff
@@ -380,3 +382,23 @@ export const fakeCollateralAsset: CollateralAsset = {
   ),
   price: 1_000_000n,
 }
+
+export const fakePreprocessedL2TransactionsStatistics = (
+  count?: number
+): PreprocessedL2TransactionsStatistics => {
+  return {
+    depositCount: count ?? randomInt(0, 100),
```
adamiak, line 390: I think it's not correct to use random values in unit tests, as they make the tests flaky and unreliable (except for things that don't matter, like fake starkKey, etc.). They're fine in fuzz testing (but there you repeat them thousands of times). Here, come up with some random values and hardcode them.

One of the reasons you should not use randoms is that in the actual test you can't compare results to the correct expected value, but you need to *implement the logic again to have something to compare to* (e.g. for unit test for this data you actually sum these random values in the test and compare to function output). So now you can have a bug in the implementation of the logic in your unit test (so maybe write a unit test for a unit test ;)). That's why the idea is that unit tests should have mostly hardcoded values and no logic.

adamiak (changes requested):

packages/backend/src/core/preprocessing/StateDetailsPreprocessor.ts
```diff
@@ -43,6 +48,56 @@ export class StateDetailsPreprocessor {
...
+
+      const previousStateUpdateDetails =
+        await this.preprocessedStateDetailsRepository.findByStateUpdateId(
+          recordToUpdate.stateUpdateId - 1,
+          trx
+        )
+
+      if (recordToUpdate.stateUpdateId > 1 && !previousStateUpdateDetails) {
+        throw new Error(
+          'Something wrong went when catching up L2 transactions for state details!'
```
adamiak, line 80: Maybe change to more explicit: Statistics for previous state update where not preprocessed. This should never happen.

packages/backend/src/core/preprocessing/UserL2TransactionsPreprocessor.ts
```diff
@@ -0,0 +1,80 @@
...
+      stateUpdateId++
+    ) {
+      const starkKeys =
+        await this.l2TransactionRepository.getStarkKeysByStateUpdateId(
+          stateUpdateId,
+          trx
+        )
+
+      for (const starkKey of starkKeys) {
+        this.logger.info(
```
adamiak, line 34: I think this log will be producing too many entries and should be removed.

packages/backend/src/peripherals/database/migrations/037_preprocessed_l2_transaction.ts
```diff
@@ -0,0 +1,47 @@
...
+      table.increments('id').primary()
+      table
+        .integer('state_update_id')
+        .notNullable()
+        .references('state_update_id')
+        .inTable('preprocessed_state_updates')
+      table.string('stark_key').notNullable()
+      table.jsonb('l2_transactions_statistics').notNullable()
+      table.jsonb('cumulative_l2_transactions_statistics').notNullable()
+      table.index(['stark_key', 'state_update_id'])
```
adamiak, line 33: The order of columns in the index is important. Since we often select by `state_update_id` only, it should come first, because if it comes second, this index won't be used (unless something changed since I learned about pg indexes :)).

BTW, we still need to spend some time afterwards and make sure we're not missing any indexes, and the ones we have are utilized and don't have too much data.

adamiak (changes requested):

packages/backend/src/peripherals/database/migrations/037_preprocessed_l2_transaction.ts
torztomasz: What if we query by stark key as often as state update id? 😅

packages/backend/src/peripherals/database/L2TransactionRepository.ts
```diff
@@ -255,14 +294,19 @@ export class L2TransactionRepository extends BaseRepository {
...
+    const knex = await this.knex(trx)
     const row = await knex('l2_transactions').where({ id }).first()
 
     return row ? toRecord(row) : undefined
   }
 
-  async findByTransactionId(id: number): Promise<AggregatedRecord | undefined> {
+  async findAggregatedByTransactionId(
+    id: number
+  ): Promise<AggregatedRecord | undefined> {
```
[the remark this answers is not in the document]
torztomasz: why if its only used in controller?

packages/backend/src/peripherals/database/L2TransactionRepository.ts
```diff
@@ -165,47 +173,78 @@ export class L2TransactionRepository extends BaseRepository {
     return parentId
   }
 
-  async countAllDistinctTransactionIds() {
+  async countByTransactionId(transactionId: number): Promise<number> {
```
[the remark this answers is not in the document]
torztomasz: same here

packages/backend/src/peripherals/database/L2TransactionRepository.ts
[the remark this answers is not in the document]
adamiak: fair point. I just prefer to add support for transactions in all new db methods, but not strictly necessary.

Btw, when I left this comment I didn't know if this function was used in preprocessor and you *forgot* to pass transaction, or simply you don't need to pass transaction. So I mostly wanted this to be double checked.

adamiak (approved):

PR #447, opened 2023-07-28
Improve UX for forced actions

torztomasz:
Resolves L2B-1989


packages/frontend/src/view/pages/user/components/UserAssetTable.tsx
```diff
@@ -93,3 +102,8 @@ export function UserAssetsTable(props: UserAssetsTableProps) {
     />
   )
 }
+
+export function getEscapeHatchLink(
+  vaultOrPositionId: string,
+  ethereumAddress: EthereumAddress
+) {}
```
adamiak, line 109: This implementation seems surprisingly short :)

torztomasz: hahah, true

adamiak (approved):

PR #448, opened 2023-07-28
Add escape related txs to database

torztomasz:
Resolves L2B-1931, L2B-1990, L2B-1978


packages/backend/src/core/TransactionHistory.ts
```diff
@@ -50,6 +50,26 @@ export class TransactionHistory {
     }
   }
 
+  getNonRevertableTransactionHistory() {
```
adamiak, line 53: What is "NonRevertable" Transaction? All transactions can be reverted, including freeze and escape. We should also show them on user page if they are reverted.

packages/backend/src/api/controllers/UserController.ts
```diff
@@ -553,9 +552,9 @@ function buildUserTransactions(
   collateralAsset?: CollateralAsset,
   assetDetailsMap?: AssetDetailsMap
 ): TransactionEntry[] {
+  const userTransactionHashes = userTransactions.map((t) => t.transactionHash)
   const sentEntries = sentTransactions
-    // Mined non-reverted transactions will be inside userTransactions
-    .filter((t) => t.mined === undefined || t.mined.reverted)
+    .filter((t) => !userTransactionHashes.includes(t.transactionHash))
```
adamiak, line 557: I think we should show "Reverted" transactions in user history. But maybe I misunderstand the purpose of this function here.

packages/backend/src/core/collectors/WithdrawalAllowedCollector.ts
```diff
@@ -70,17 +74,34 @@ export class WithdrawalAllowedCollector {
...
+        const freezeStatus = await this.keyValueStore.findByKey('freezeStatus')
+
+        const data = {
+          starkKey: StarkKey.from(event.args.starkKey),
+          assetType: AssetHash.from(event.args.assetType),
+          nonQuantizedAmount: event.args.nonQuantizedAmount.toBigInt(),
+          quantizedAmount: event.args.quantizedAmount.toBigInt(),
+        }
+
+        if (freezeStatus === 'frozen') {
```
adamiak, line 87: I guess this is a bit confusing, but freezeStatus is the status of the contract right now (it's set by calling `.isFrozen()` on the contract). So when explorer is resyncing, it might be processing LogWithdrawalEvents from the past, when the exchange was not frozen yet. So checking freezeStatus here is incorrect.

You can even imagine a scenario where freeze comes exactly after this event, so relying on anything that is set in a step which is separate to the collection process (e.g. before or after) could also return incorrect value here.

I think the best solution would be to add collector for an event that's fired when exchange is frozen (`emit LogFrozen()`)and here check if this event is in DB.

packages/frontend/src/view/components/tables/TransactionsTable.tsx
```diff
@@ -108,7 +114,9 @@ function getStatus(transaction: TransactionEntry): {
       case 'REVERTED':
         return { type: 'ERROR', text: 'REVERTED' }
       case 'INCLUDED':
-        throw new Error('WITHDRAW transaction cannot be INCLUDED')
+        throw new Error(
+          `${transaction.status} transaction cannot be ${transaction.status}`
```
adamiak, line 118: I don't understand this log message :)

packages/backend/src/api/controllers/TransactionSubmitController.ts
```diff
@@ -144,50 +141,172 @@ export class TransactionSubmitController {
...
+    const timestamp = Timestamp.now()
+    const fetched = await this.fetchTxAndDecode(
+      transactionHash,
+      this.contracts.escapeVerifier,
+      validateVerifyEscapeRequest
+    )
+    if (!fetched.isSuccess) {
+      return fetched.controllerResult
+    }
+
```
adamiak, line 228: You still need to check if fetched.data is true or false

adamiak (changes requested):

packages/backend/src/peripherals/database/transactions/UserTransaction.ts
```diff
@@ -21,8 +21,9 @@ export type UserTransactionData =
   | ForcedTradeData
   | ForcedWithdrawalData
   | FullWithdrawalData
-  | EscapeVerifiedData
   | WithdrawalPerformedData
+  | EscapeVerifiedData
```
adamiak, line 25: It turns out that parameters to `verifyEscape()` and `escape()` are different for spot and perpetual, even though the whole process is the same.

Perpetual:
```    
function verifyEscape(
        uint256[] calldata merkleProof,
        uint256 nAssets,
        uint256[] calldata sharedState
    )
    
 function escape(
        uint256 starkKey,
        uint256 vaultId,   // <-- this acutally is a positionId
        uint256 quantizedAmount
    )
 ```

Spot:
    
```
function verifyEscape(
          uint256[] calldata escapeProof
)     

function escape(
        uint256 ownerKey,
        uint256 vaultId,
        uint256 assetId,
        uint256 quantizedAmount
    )
```

Because of that, and to make things consistent, I suggest renaming the UserTransaction types and SentTransaction types to:

* VerifyPerpetualEscape
* FinalizePerpetualEscape

* VerifySpotEscape
* FinalizeSpotEscape

and to modify all function names accordingly (e.g. encode... etc.)

torztomasz: Well, I agree we need that refactoring, although I would do this as a part of spot escapes PR

packages/backend/src/api/controllers/TransactionSubmitController.ts
torztomasz: No need, in verifyAndDecodeTx fn there is a check for this: !data

packages/backend/src/api/controllers/UserController.ts
torztomasz: revist after log frozen

adamiak (approved):

PR #468, opened 2023-08-18
Update statistics design

torztomasz:


packages/frontend/src/view/pages/home/components/HomeStatistics.tsx
```diff
@@ -76,12 +75,25 @@ function StatisticsItem({
...
+        {formatStatisticsCount(value)}
+      </span>
+      <span className="hidden text-sm font-semibold text-zinc-500 md:inline">
+        # of all current {title.toLowerCase()}
+      </span>
     </div>
   )
 }
+
+function formatStatisticsCount(count: number) {
```
sz-piotr, line 93: We could use the number formatting function we have in L2BEAT, so that we display nicer numbers overall.

sz-piotr (approved):

PR #482, opened 2023-08-25
Implement tutorials page

torztomasz:
Resolves L2B-2274


packages/backend/src/core/TutorialService.ts
```diff
@@ -0,0 +1,17 @@
+import { HomeTutorialEntry } from '@explorer/frontend'
+import fs from 'fs'
+
+export class TutorialService {
+  getTutorials(): HomeTutorialEntry[] {
+    const files = fs.readdirSync('src/content/tutorials')
```
sz-piotr, line 6: Will the tutorials always be platform neutral? So the same for e.g. dydx and apex?

packages/frontend/src/preview/routes.ts
```diff
@@ -131,6 +132,7 @@ const routes: Route[] = [
 
       ctx.body = renderHomePage({
         context,
+        tutorials: tutorials,
```
sz-piotr, line 135:
```suggestion
        tutorials,
```
Same in cases below

sz-piotr (approved):

packages/backend/src/core/TutorialService.ts
torztomasz: Well I would say so, if not we would filter by some prefix e.g. apex-example.md

PR #485, opened 2023-08-28
Ignore but log Live L2 txs parse errors

adamiak:


packages/backend/src/peripherals/starkware/schema/PerpetualLiveL2TransactionResponse.ts
```diff
@@ -5,14 +5,19 @@ import { PerpetualL2Transaction } from './PerpetualBatchInfoResponse'
 const PerpetualLiveL2TransactionResponseTransactionInfo = z.object({
   tx: PerpetualL2Transaction,
   tx_id: z.number(),
+  parseError: z.undefined(),
```
torztomasz, line 8: why is that z.undefined?

torztomasz (dismissed):

packages/backend/src/peripherals/starkware/schema/PerpetualLiveL2TransactionResponse.ts
adamiak: Removed here and added as a separate discriminated union called ParsedL2TransactionInfo

packages/backend/src/core/sync/LiveL2TransactionDownloader.test.ts
```diff
@@ -86,17 +87,19 @@ describe(LiveL2TransactionDownloader.name, () => {
...
-          ).toHaveBeenNthCalledWith(
-            i + 1,
-            {
-              transactionId: tx.transactionId,
-              timestamp: tx.timestamp,
-              data: tx.transaction,
-            },
-            mockKnexTransaction
-          )
+          if (tx.parsedSuccessfully) {
```
michalsidzej, line 90: I'd suggest adding a test that checks what happens if the transaction is not parsed successfully.

michalsidzej (approved):
This change is quite simple, but touches quite a few logical checks. Would be great to unit test a less happy path. This is a nit pick, but otherwise LGTM

packages/backend/src/core/sync/LiveL2TransactionDownloader.test.ts
adamiak: Added

PR #499, opened 2023-11-09
Remove spotlight article

adamiak:
The size of the spotlight component is too big for the content, and we need to rethink if it even makes sense to show the tutorial there.

### DIFF
--- a/packages/frontend/src/view/pages/home/HomePage.tsx
+++ b/packages/frontend/src/view/pages/home/HomePage.tsx
@@ -71,7 +71,7 @@ function HomePage(props: HomePageProps) {
             />
             {props.tutorials.length > 0 && (
               <HomeTutorials
-                tutorials={props.tutorials.slice(0, MAX_TUTORIALS )}
+                tutorials={props.tutorials.slice(0, MAX_TUTORIALS)}
                 showViewAll={showViewAllTutorials}
                 className="hidden xl:flex"
               />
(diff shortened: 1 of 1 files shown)


packages/frontend/src/view/pages/home/HomePage.tsx
```diff
@@ -87,12 +85,6 @@ function HomePage(props: HomePageProps) {
               className="xl:hidden"
             />
           )}
-          {lastTutorial && (
```
torztomasz, line 90: lets also change MAX_TUTORIALS to 2 then on line 76 remove the "- 1" part and change 48 lane to `const showViewAllTutorials = props.tutorials.length > MAX_TUTORIALS`

torztomasz (changes requested):

packages/frontend/src/view/pages/home/HomePage.tsx
```diff
@@ -42,11 +41,10 @@ export function renderHomePage(props: HomePageProps) {
   return reactToHtml(<HomePage {...props} />)
 }
 
-const MAX_TUTORIALS = 3
+const MAX_TUTORIALS = 2
 
 function HomePage(props: HomePageProps) {
-  const showViewAllTutorials = props.tutorials.length > 3
-  const lastTutorial = props.tutorials[MAX_TUTORIALS - 1]
+  const showViewAllTutorials = props.tutorials.length > 2
```
torztomasz, line 47:
```suggestion
  const showViewAllTutorials = props.tutorials.length > MAX_TUTORIALS
```

torztomasz (approved):

PR #540, opened 2026-02-16
Add IP Rate Limitting Middleware

adamiak:
Allow to limit number of requests per minut by IP, by setting IP_RATE_LIMIT_PER_MINUTE env variable.


torztomasz (approved):
code looks good overall
however why not just smash cloudflare at it? that's what we have done for api.l2beat.com and some more

